PR #8, opened 2018-09-26
Added PSNR comparison metric to reducer tool and WebUI

paulthomson:


shadersets-util/src/main/java/com/graphicsfuzz/shadersets/MetricImageFileComparator.java
```diff
@@ -48,8 +65,8 @@ public boolean areFilesInteresting(File reference, File variant) {
       LOGGER.info(": difference is " + diff);
       return result;
     } catch (IOException exception) {
-      LOGGER.info("Not interesting - exception occurred during histogram comparison.");
-      return false;
+      LOGGER.error("Exception occurred during histogram comparison.", exception);
```
hevrard: In log message: "histogram comparison" should be "image comparison", would be great to add the used metric name if possible?

PR #15, opened 2018-09-27
Shader job

afd:
A big refactoring to the way that shaders are handled during generation, and better support for generating Vulkan-ready shaders.


paulthomson:
Looks good but does not build.

paulthomson:
Actually, it is possible that master does not build or it is just my issue. I will check.

PR #36, opened 2018-10-03
Add vulkan worker source

hevrard:
I tried to cut off unnecessary files and folders from the original vulkan-samples repo, but it's still a big amount of code landing. You should be able to open third_party/vulkan-worker/src as a gradle project in Android studio, and build the vulkan_worker module. You need latest Android Studio v3.2.0.


hevrard:
Closing the pull request for now.

PR #39, opened 2018-10-05
Rework run shader set

afd:
Made run shader set work on a shader job prefix, rather than a .frag shader, and did some renaming of shader set to shader family through the code base (there is still a bit of shader set left; no urgency to fix it everywhere.)


afd:
I overrode the CLA status because the edits were by Hugues Evrard, who is a Googler.

PR #49, opened 2018-10-08
Better spirv opt

hevrard:
Catch spirv-opt bad behaviors:
- compilation errors
- timeouts (30sec)


hevrard:
Actually the spirv-opt timeout is 10 seconds. Let me know if you think it should be different :)

PR #51, opened 2018-10-10
Added script to find minimal set of spirv-opt flags that triggers an assertion failure

afd:
Added script to find minimal set of spirv-opt flags that triggers an assertion failure.


paulthomson:
Needs license.

PR #56, opened 2018-10-10
Improve webui

### LINKED ISSUE #29: WebUI: reduce size of server log embedded in home page
With daily snapshot of logs and long experiments, the current log can reach 10MB, which makes the home page too long to load. A solution would be to only load last 1000 lines of the log for instance.

hevrard:
Fix #29


paulthomson:
Rebased. Resolved conflicts.

PR #59, opened 2018-10-11
Squashed ShaderJobFile refactor.

paulthomson:


paulthomson:
**ShaderJobFileOperations**

The main idea with this PR is to move all assumptions about the locations of files into `ShaderJobFileOperations`. Existing tests and the `WebUi` still make assumptions. The rest of the code should only refer to the following (as much as possible):

* File `shaderJobFile` e.g. variant.json (alongside: variant.frag, variant.vert, etc.)
* File `shaderJobResultFile` e.g. variant.info.json (alongside: variant.png, variant.txt, etc.)
* `ShaderJob`: contains the TranslationUnits of fragment and vertex shaders, uniform info, license.
* `ImageJob`: the contents of a shaderJobFile in memory (Thrift object), plus other options about how to run it.
* `ImageJobResult`: the contents of a shaderJobResultFile in memory (Thrift object), plus extra info.

Using the methods in `ShaderJobFileOperations`, you can copy, move, delete, check if exists on the files, and convert back and forth between most of the above "types". 

In fact, nearly all file operations (even simple operations like `moveFile`, `mkdir`) are also now implemented in `ShaderJobFileOperations` (so the name is not exactly appropriate); this allows mocking out the file system, although perhaps it is more useful for quickly understanding which file system operations we use (and where) and ensuring that we always implement file system operations consistently (e.g. using Files. or FileUtils., etc.). 

**Compute shaders**: I have removed CompteJob, and merged it into ImageJob, so ImageJob is no longer a good name; ImageJob is now essentially a "ShaderJob" or a "GPUJob". In future, it could in theory have many shaders of various different types in a single job; the framework doesn't need to care.

PR #157, opened 2018-12-05
Be more resilient when dumping worker info

hevrard:


hevrard:
Thanks for the feedback, I added a variable and a comment, but now it needs to be reviewed once more.

PR #163, opened 2018-12-11
Vulkan worker: use a STARTED file to detect whether the app has started

paulthomson:
I have not tested this yet, but see what you think of this approach.


python/src/main/python/drivers/vkrun.py
```diff
@@ -139,20 +139,26 @@ def run_android(vert, frag, json, skip_render):
...
 
-        # Begin the busy-wait loop by sleeping to let the app start
-        # properly. The 'adb shell am start' may return before the app is
-        # actually started (this is all asynchronous), in which case we may not
-        # be able to detect the app pid.
-        time.sleep(0.1)
+        # Don't pass here until app has started.
+        if status == 'UNEXPECTED_ERROR':
+            if adb('shell test -f ' + ANDROID_SDCARD + '/STARTED').returncode != 0:
+                continue
```
hevrard: I suggest to add a time.sleep(0.1) while waiting for the STARTED file, otherwise it is busy-looping without any sleep.

paulthomson: OK done.

PR #173, opened 2018-12-14
DEV: Rework shader generation in Java

afd:
Shader generation is now written in Java, rather than Python, reducing the extent to which Python is a dependency and making testing of shader generation simpler as it can be done in JUnit.

- Fixed an issue where do ... while loops were generated in WebGL 1.0 shaders.

- Fixed an infinite loop issue in the generator, where if no available transformation could make progress the transformations would be applied forever.

- Added a workaround for a memory exhaustion issue in shader translator.

- Added tests for shader family generation.

- Changed generation top-level script so that - is used instead of _ to separate arguments, in line with the way glsl-reduce works.  Also changed the order of arguments so that the references are specified before the donors.  Removed the need for the user to explicitly make the output directory for shaders.


generator/src/main/java/com/graphicsfuzz/generator/tool/GenerateShaderFamily.java
```diff
@@ -0,0 +1,363 @@
...
+    // Validate reference shaders.
+    if (!disableValidator) {
+      if (!(fileOps.areShadersValid(preparedReferenceShaderJob, false)
+          && fileOps.areShadersValidShaderTranslator(preparedReferenceShaderJob, false))) {
+        throw new RuntimeException("One or more of the prepared shaders of shader job "
+            + preparedReferenceShaderJob.getAbsolutePath() + " is not valid.");
+      }
+    }
+
+    if (primitivesFile(referenceShaderJob).isFile()) {
```
paulthomson: Maybe can use: ShaderJobFileOperations.copyShaderJobFileTo.
And just generally: no direct file operations if possible.

generator/src/main/java/com/graphicsfuzz/generator/tool/GlslGenerate.java
```diff
@@ -0,0 +1,175 @@
...
+    final File donorsDir = ns.get("donors");
+    final File outputDir = ns.get("output_dir");
+    final String prefix = ns.get("prefix");
+    final int numVariants = ns.getInt("num_variants");
+    final boolean verbose = ns.getBoolean("verbose");
+    final int seed = ns.getInt("seed");
+
+    FileUtils.forceMkdir(outputDir);
+
+    final File[] referenceShaderJobFiles = referencesDir.listFiles((dir, name) -> name.endsWith(
```
paulthomson: there is probably a method for this

paulthomson: listShaderJobFiles

afd: Done.

generator/src/main/java/com/graphicsfuzz/generator/tool/GenerateShaderFamily.java
afd: Leaving as is for now, since primitives is a bit complex and we might get rid of it anyway.

PR #174, opened 2018-12-14
vulkan worker: python robustness changes

paulthomson:


python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -209,51 +217,54 @@ def doImageJob(args, imageJob):
...
 
 ################################################################################
 
-def isDeviceAvailable(serial):
+
+def is_device_available(serial):
     cmd = 'adb devices'
-    devices = subprocess.run(cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE, timeout=2).stdout.splitlines()
+    devices = subprocess.run(cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE,
+                             timeout=2).stdout.splitlines()
```
afd: Should we use a larger timeout here?  If the timeout is exceeded the worker quits, so timing out is kind of a last resort.  So maybe use 5?  Also could we use a named constant for this purpose?

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -99,13 +108,13 @@ def prepare_shaders(frag_file, frag_spv_file, vert_spv_file):
...
-
 ################################################################################
 
-def doImageJob(args, imageJob):
-    name = imageJob.name.replace('.frag','')
-    fragFile = name + '.frag'
-    jsonFile = name + '.json'
+
+def do_image_job(args, image_job):
+    name = image_job.name[:-5]
```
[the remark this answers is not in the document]
afd: I meant the above to be a review comment.

python/src/main/python/drivers/vkrun.py
```diff
@@ -15,67 +15,78 @@
...
-import shutil
+from typing import Union, IO, Any
 
 ################################################################################
 # Constants
 
+
 LOGFILE = 'vklog.txt'
 TIMEOUT_RUN = 30
 NUM_RENDER = 3
```
afd: I like the above and it would be good to do the same for other constants (e.g. timeout)

python/src/main/python/drivers/vkrun.py
```diff
@@ -15,67 +15,78 @@
...
     output = ''
-    if shader[-5:] == '.frag' or shader[-5:] == '.vert':
+    if shader.endswith('.frag') or shader.endswith('.vert'):
         output = shader + '.spv'
         cmd = 'glslangValidator -V ' + shader + ' -o ' + output
         subprocess.check_call(cmd, shell=True, timeout=TIMEOUT_RUN)
-    elif shader[-9:] == '.frag.asm' or shader[-9:] == '.vert.asm':
-        output = shader.replace('.asm', '.spv')
+    elif shader.endswith('.frag.asm') or shader.endswith('.vert.asm'):
+        output = shader[:-4] + '.spv'
```
afd: Can we avoid the explicit use of 4 here, and use len('.spv')?

python/src/main/python/drivers/vkrun.py
```diff
@@ -15,67 +15,78 @@
...
         cmd = 'spirv-as ' + shader + ' -o ' + output
         subprocess.check_call(cmd, shell=True, timeout=TIMEOUT_RUN)
-    else:
-        assert(shader[-4:] == '.spv')
+    elif shader[-4:] == '.spv':
```
afd: Similar comment here.

docs/glsl-fuzz-walkthrough.md
```diff
@@ -603,6 +612,48 @@ the reduction of this variant leads to the following files:
...
+to run SPIR-V shaders on the device.
+You can use this script directly
+if you want to re-run some SPIR-V shaders.
+Your device must show in `adb devices`.
+For example:
+
+```sh
+vkrun test.vert variant_005.frag variant_005.json --serial ABCD
+```
+
```
afd: It's a little confusing above: we say that vkrun enables running SPIR-V shaders on the device, but then the example of how to use vkrun involves a .frag file.  We should say that you can use this to run a single .frag shader on a Vulkan device, via translation to .spv, and make it clear that the convenience here over run-shader-family is that it does not require a server.

Does vkrun have an option where you can give it a .spv file instead of a .frag file?  If so it would be good to document that too.

docs/glsl-fuzz-walkthrough.md
```diff
@@ -375,6 +375,17 @@ the `vulkan-worker-android` app running on the Android device.
...
+For example:
+
+```
+test.vert             # The default GLSL vertex shader.
+variant_005.json      # The shader job file (containing uniforms data).
+variant_005.frag      # The GLSL fragment shader.
+variant_005.frag.spv  # The SPIR-V version of the fragment shader.
+variant_005.vert.spv  # The SPIR-V version of the default vertex shader.
+```
+
```
afd: If spirv-opt is invoked within the Vulkan worker do we get .opt.spv files (or something similar) as well?  It would be good if we did and would be good to document that.

docs/glsl-fuzz-walkthrough.md
```diff
@@ -562,6 +562,19 @@ In the above example,
...
+## After updating a device
+
+If you update a device,
+its "fingerprint" will change
+and the server will usually reject it.
+The solution is to give the worker a new name.
+This also allows you to compare
+results across different versions of your device
+in the Web UI,
+without having to delete your old results.
```
afd: We should perhaps acknowledge that if you want to e.g. run reductions on already computed results, but that your worker info has changed, you can't just rename the worker.

We should perhaps provide override functionality for this.

paulthomson: Perhaps, although I would argue that you would want to first check if the bugs reproduces?

PR #207, opened 2019-01-18
Automatically restrict shader colours

afd:
This adds to GenerateAndRunShaders the ability to restrict the colours a shader can produce to a very small, known palette.  This makes it possible to find bugs by checking for cases where inclusion in that restricted palette is violated.

GenerateAndRunShaders is an experimental part of the repo - it's included in our releases but not yet documented.  This addition is similarly rather experimental.  We will polish it up if it turns out to be promising for finding bugs.


paulthomson:
It looks like this addition is not optional, but I am not certain. Just wanted to check.

PR #210, opened 2019-01-21
Fix documentation of generator

hevrard:


docs/glsl-fuzz-generate.md
```diff
@@ -115,7 +115,7 @@ Assuming you have extracted the `graphicsfuzz.zip` file to get `graphicsfuzz/`:
 
 ```sh
 # Copy the sample shaders into the current directory:
-cp -r graphicsfuzz/shaders/samples samples
+cp -r graphicsfuzz/shaders/src/main/glsl/samples samples
```
paulthomson: This is the path from the release zip, so it should be correct, but perhaps we should clarify

afd:
I had to make changes to the generator documentation as part of #216, so have replicated the  changes from this PR in that PR.  Closing this one.

PR #211, opened 2019-01-21
Add compute shader samples

hevrard:


shaders/src/main/glsl/samples/compute/310es/comp-0001-findmax.comp
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2018 The GraphicsFuzz Project Authors
```
paulthomson: Copyright year to 2019

PR #218, opened 2019-01-22
gapidfuzz: add optional support for .vert and .comp shaders

paulthomson:


paulthomson:
Yes it is a bit of an experiment, but it is certainly easier for the programmer at this stage, and I get type hints.

PR #227, opened 2019-01-29
Summer of Code ideas

afd:


afd:
Overriding our protection since this is not a code-related change, and we want the Summer of Code ideas to be visible on master.

PR #232, opened 2019-01-29
Fix gapidfuzz test and avoid painful characters in dumped shader filenames

hevrard:


hevrard:
> I think the shader filename is used later when replacing the shader in the trace. Need to check if that will still work. The filename is the “shader handle” as used by gapid.

Yeah good point, I reverted that change in new commits!

PR #242, opened 2019-02-04
Fixed pretty-printing of for loop with empty header, and removed some…

afd:
… assumptions about available precision that were being made when 'injectionSwitch' or structs were added.


afd:
Starting again with this one as I rebased wrongly.

PR #262, opened 2019-02-13
Add alpha channel support for histogram comparison

### LINKED ISSUE #164: Use image comparison metric that considers alpha channel
The image histogram and PSNR metrics that GraphicsFuzz uses do not consider the alpha channel.  They thus miss shader compiler bugs that lead to incorrect alpha computation.

We should consider ways to adjust these metrics so that they do consider alpha (or look at different metrics).

paulthomson:
Also update opencv. This fixes #164, although it would still be good to consider other image comparison algorithms.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/ImageUtil.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/ImageUtil.java
@@ -65,7 +65,7 @@ public class ImageUtil {
     }
 
     // matWithAlpha has four channels.
-    assert matWithAlpha.type() != opencv_core.CV_8UC4;
+    assert matWithAlpha.type() == opencv_core.CV_8UC4;
 
     opencv_core.Mat mat = new opencv_core.Mat();
     // Remove alpha so we can convert colors to HSV.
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/graphicsfuzz/common/util/ImageUtil.java
```diff
@@ -46,16 +48,60 @@
...
+      // The file could be in several different formats, including greyscale.
+      // Reload the image, converting to BGR.
+      matWithAlpha = opencv_imgcodecs.imread(file, opencv_imgcodecs.IMREAD_COLOR);
+      assert matWithAlpha.type() == opencv_core.CV_8UC3;
+      // Add (opaque) alpha channel.
+      opencv_imgproc.cvtColor(matWithAlpha, matWithAlpha, opencv_imgproc.COLOR_BGR2BGRA);
+    }
+
+    // matWithAlpha has four channels.
+    assert matWithAlpha.type() != opencv_core.CV_8UC4;
```
hevrard: Comment not in sync with assertion?

paulthomson: Yes. Thanks!

PR #263, opened 2019-02-14
Integration of more code to do semantics-changing mutations

afd:
This integrates code that was developed as part of a Chrome University Research Project (with IP owned by Google) for performing semantics-changing mutations on a shader.

The mutations do things like replace a 2-argument binary operator with a function call that takes two arguments.


generator/src/main/java/com/graphicsfuzz/generator/semanticschanging/InterchangeExprMutationFinder.java
```diff
@@ -0,0 +1,164 @@
...
+  private final IRandom generator;
+
+  public InterchangeExprMutationFinder(TranslationUnit tu,
+                                       IRandom generator) {
+    super(tu);
+    this.generator = generator;
+  }
+
+  @Override
+  protected void visitExpr(Expr expr) {
```
paulthomson: As discussed, it looks like this is overriding a generic "visitExpr" visitor method and then giving up on this expression and all its children when the type cannot be identified, even though the type of the children could be know.

afd: Thanks for pointing this out.  I just filed #264 which explains why visitExpr is ill-named.  I will do a separate PR to rename it, as that has nothing to do with this PR.

generator/src/test/java/com/graphicsfuzz/generator/semanticschanging/InterchangeExprMutationFinderTest.java
```diff
@@ -0,0 +1,50 @@
...
+public class InterchangeExprMutationFinderTest {
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void testValidInterchanges() throws Exception {
+    final IRandom generator = new RandomWrapper(0);
+    final ShaderJobFileOperations fileOperations = new ShaderJobFileOperations();
+    for (File shaderJobFile : Paths.get(ToolPaths.getShadersDirectory(), "samples",
```
paulthomson: Technically there is `listShaderJobFiles` for this. Not sure how important this is though. It ensures they are sorted and so makes things more deterministic.

afd: Done, thanks!

PR #275, opened 2019-02-24
Amber-based worker, and tests for SPIR-V runner

afd:
Added tests for SPIR-V runner, and used these tests to refactor the worker and runner so that both Amber and the legacy runner are supported in the glsl-to-spv worker.


python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+
+def prepare_device(wait_for_screen: bool, using_legacy_worker: bool):
+    adb_check('logcat -c')
+
+    if using_legacy_worker:
+        # If the legacy worker is being used, give it the right permissions, stop it if already running, and get its
+        # working directory ready.
+        adb_check('shell pm grant com.graphicsfuzz.vkworker android.permission.READ_EXTERNAL_STORAGE')
+        adb_check('shell pm grant com.graphicsfuzz.vkworker android.permission.WRITE_EXTERNAL_STORAGE')
+        adb_can_fail('shell am force-stop ' + ANDROID_LEGACY_APP)
```
hevrard: Force-stopping the worker everytime is good to make sure it is starting in a clean state, but I just want to point out that it is not strictly required and is likely to have performance impacts.

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+    # If the app never starts, status remains as UNEXPECTED_ERROR.
+    # Once the app starts, status becomes TIMEOUT.
+    # If we reach the end of the loop below, we break and the status is updated just before.
+
+    while time.time() < deadline:
+        time.sleep(BUSY_WAIT_SLEEP_SLOW)
+
+        # Don't pass here until app has started.
+        if status == 'UNEXPECTED_ERROR':
+            if adb_can_fail('shell test -f ' + ANDROID_SDCARD_GRAPHICSFUZZ_DIR + '/STARTED').returncode != 0:
```
hevrard: All adb_can_fail() commands may raise a subprocess.TimeoutExpired exception, which I guess we will want to catch in such busy-waiting loops.

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+
+    with open(statusfile, 'w') as f:
+        f.write(status)
+
+    if status != 'SUCCESS':
+        # Something went wrong. Make sure we stop the app.
+        adb_can_fail('shell am force-stop ' + ANDROID_LEGACY_APP)
+
+    # Grab image if present.
+    image_path = ANDROID_SDCARD_GRAPHICSFUZZ_DIR + '/image_0.png'
```
hevrard: The file has already be retrieved on line 319, so we may drop this image file grabbing.

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+        return prefix + '.' + kind
+    if os.path.isfile(prefix + '.' + kind + '.asm'):
+        return prefix + '.' + kind + '.asm'
+    assert os.path.isfile(prefix + '.' + kind + '.spv')
+    return prefix + '.' + kind + '.spv'
+
+
+def main_helper(args):
+    # TODO: update description to reflect final output.
+    description = 'Run SPIR-V shaders.  Output: ' + LOGFILE_NAME + ', image.png'
```
hevrard: Should we document compute shader output?

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -288,18 +298,19 @@ def do_compute_job(args, comp_job):
...
+        with open(runspv.LOGFILE_NAME, 'r', encoding='utf-8', errors='ignore') as f:
             res.log += f.read()
 
     if os.path.isfile('STATUS'):
         with open('STATUS', 'r') as f:
             status = f.read().rstrip()
         if status == 'SUCCESS':
             res.status = tt.JobStatus.SUCCESS
-            assert(os.path.isfile('ssbo.json'))
+            assert (os.path.isfile('ssbo.json'))
```
paulthomson: Nit: assert does not need parens. Also can be fixed elsewhere.

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -120,22 +125,22 @@ def glsl2spv(glsl, spv):
...
+
 ################################################################################
 
 
 def run_spirv_opt(spv_file, args):
     print("Running optimizer.")
     spv_file_opt = spv_file + '.opt'
     cmd = os.path.dirname(HERE) + '/../../bin/' + get_bin_type() + '/spirv-opt ' + \
-        args.spirvopt + ' ' + spv_file + ' -o ' + spv_file_opt
+          args.spirvopt + ' ' + spv_file + ' -o ' + spv_file_opt
```
paulthomson: This now shows a PEP 8 (i.e. style) failure and there are others throughout this file. It would be good to fix them all and/or decide if we are out-of-sync somehow.

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -288,18 +298,19 @@ def do_compute_job(args, comp_job):
             res.status = tt.JobStatus.UNEXPECTED_ERROR
             return res
 
-    vkrun.run_compute(tmpcompspv, tmpjson)
+    assert not args.legacy_worker
+    runspv.run_compute_amber(comp=tmpcompspv, json=tmpjson, output_dir=os.getcwd(), force=args.force)
```
paulthomson: PEP 8 failure: `is_android` parameter is missing.

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -216,18 +222,21 @@ def do_image_job(args, image_job):
...
-    remove(vkrun.LOGFILE)
+    remove(runspv.LOGFILE_NAME)
 
-    if args.linux:
-        vkrun.run_linux(vert_spv_file, frag_spv_file, json_file, skip_render)
-    elif args.vkrunner:
-        vkrun.run_vkrunner(vert_spv_file, frag_spv_file, json_file)
+    if args.legacy_worker:
+        if args.target == 'host':
+            runspv.run_image_android_legacy(vert=vert_spv_file, frag=frag_spv_file, json=json_file,
```
paulthomson: This should be `run_image_host_legacy`?

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -216,18 +222,21 @@ def do_image_job(args, image_job):
...
+                                            output_dir=os.getcwd(), force=args.force, skip_render=skip_render)
+        else:
+            runspv.run_image_android_legacy(vert=vert_spv_file, frag=frag_spv_file, json=json_file,
+                                            output_dir=os.getcwd(), force=args.force, skip_render=skip_render)
     else:
-        wait_for_screen = not args.force
-        vkrun.run_android(vert_spv_file, frag_spv_file, json_file, skip_render, wait_for_screen)
+        runspv.run_image_amber(vert=vert_spv_file, frag=frag_spv_file, json=json_file, output_dir=os.getcwd(),
+                               force=args.force, is_android=(not args.target == 'host'))
```
paulthomson: I feel this would be better `args.target == 'android'`

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -451,30 +462,24 @@ def main():
     worker_info_file = 'worker_info.json'
     remove(worker_info_file)
 
-    if args.linux:
-        vkrun.dump_info_linux()
+    if is_android:
+        runspv.dump_info_android_legacy(wait_for_screen=not args.force)
```
paulthomson: These lines would require that, if the user is not using the legacy worker option then the user must have both the legacy worker and amber ready-to-go. Should we perhaps not require this? E.g. only dump the info when using the legacy option?

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+BUSY_WAIT_SLEEP_FAST = 0.1
+
+################################################################################
+# Common
+
+orig_print = print
+
+
+# noinspection PyShadowingBuiltins
+def print(s):
```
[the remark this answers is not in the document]
paulthomson: Yes this sounds good, but if we do this, then we should somehow disable `print` from being used in this file anyway, as we really don't want the original `print` to creep in; it does not work well on Windows

python/src/main/python/drivers/runspv.py
paulthomson: I would vote to keep this because this is how it was before and it is our legacy worker anyway.

python/src/main/python/drivers/runspv.py
paulthomson: Yeah it is hard to know what do. If we get a timeout after 30 seconds running an adb command then something has gone very wrong. We could catch this and log an "unexpected error", so that the worker is likely to continue running. However, the worker is likely to keep returning unexpected errors from this point forward, polluting all future results. If we don't catch it, the exception will propagate to the glsl-to-spv worker (because we call runspv via Python), and the worker will likely terminate. This might be the right thing to do because the Android device is in a weird state and needs to be investigated, and all future tests are probably going to fail anyway.

python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -451,30 +462,24 @@ def main():
...
 
     if not os.path.isfile(worker_info_file):
-        print('Failed to retrieve worker information. '
-              'Make sure the app permission to write to external storage is enabled.')
-        exit(1)
+        raise Exception('Failed to retrieve worker information.  If targeting Android, make sure the app permission '
+                        'to write to external storage is enabled.')
 
-    # noinspection PyUnusedLocal
-    worker_info_json_string = '{}'  # Default value: dummy but valid JSON string
```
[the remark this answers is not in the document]
paulthomson: This sounds like a good idea (and I am guessing I added `noinspection PyUnusedLocal` to ensure there is no warning for this statement), but now that I look more closely, I am guessing that  'worker_info_json_string' will either be overwritten or an exception will be thrown. If we catch exceptions here, then this would make sense and the warning will probably go away.

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: OK we might be out of sync. At least, my line length is 100, which I think is inherited from the root project, but this might not apply if you are using PyCharm from within the `python` directory.

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+        # App has terminated and there is no DONE file; this definitely looks like a crash.
+        status = 'CRASH'
+        break
+
+    # Grab log:
+    with open(logfile, 'w', encoding='utf-8', errors='ignore') as f:
+        adb_check('logcat -d', stdout=f)
+
+    # retrieve all files
+    adb_check('pull ' + ANDROID_SDCARD_GRAPHICSFUZZ_DIR + ' ' + output_dir)
```
paulthomson: I feel like we may need a / here

python/src/main/python/drivers/runspv.py
```diff
@@ -0,0 +1,944 @@
...
+
+    # Check sanity:
+    if status == 'SUCCESS':
+        sanity_before = output_dir + '/sanity_before.png'
```
paulthomson: "/" should not be used (except on the Android device). I will try to replace these.

python/src/main/python/drivers/glsl-to-spv-worker.py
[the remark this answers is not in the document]
paulthomson: OK I have changed this back and we also now catch and ignore exceptions on the file read.

python/src/main/python/test_scripts/test_runspv/README.md
```diff
@@ -0,0 +1,20 @@
+`run_spvtests.py` runs a series of tests that check `runspv` is behaving
```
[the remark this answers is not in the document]
paulthomson: Fixed

python/src/main/python/drivers/runspv.py
paulthomson: I vote to keep this

python/src/main/python/drivers/runspv.py
paulthomson: Nice. I removed this.

python/src/main/python/drivers/runspv.py
paulthomson: I made the output png and updated the description here to include image_0.png and ssbo.json.

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: I have fixed these

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: I have fixed these

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: Fixed this

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: I fixed this but it would be good to double check

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: I changed this

python/src/main/python/drivers/runspv.py
paulthomson: I replaced these

python/src/main/python/drivers/runspv.py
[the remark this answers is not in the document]
paulthomson: I vote to leave this.

python/src/main/python/drivers/glsl-to-spv-worker.py
paulthomson: I have changed this to ignore any errors when not using the legacy worker.

python/src/main/python/drivers/glsl-to-spv-worker.py
hevrard: Looks good

PR #280, opened 2019-02-26
Changed BoolConstantExpr so that instead of having two static instanc…

afd:
…es, TRUE and FALSE, we create separate objects for each instance of 'true' and 'false' in the AST.  This fixes a bug where mutating 'true' in one place was leading to a 'true' in a different place being modified.  In the process, added a check during the building of a parent map to ensure that (except in documented special cases) a node does not have multiple parents.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/visitors/StandardVisitor.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/visitors/StandardVisitor.java
@@ -315,6 +315,7 @@ public abstract class StandardVisitor implements IAstVisitor {
   public void visitParameterDecl(ParameterDecl parameterDecl) {
     visitChildFromParent(parameterDecl.getType(), parameterDecl);
     if (parameterDecl.getArrayInfo() != null) {
+      // Only visit the parameter's array information if there actually is array information.
       visitChildFromParent(this::visitArrayInfo, parameterDecl.getArrayInfo(), parameterDecl);
     }
   }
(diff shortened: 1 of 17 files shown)


common/src/main/java/com/graphicsfuzz/common/util/CannedRandom.java
```diff
@@ -67,7 +67,7 @@ public boolean nextBoolean() {
 
   @Override
   public IRandom spawnChild() {
-    throw new UnsupportedOperationException("Child spawning not available");
```
paulthomson: This is probably fine but did you mean to commit this?

ast/src/main/java/com/graphicsfuzz/common/ast/visitors/StandardVisitor.java
```diff
@@ -314,7 +314,9 @@ public void visitVariableDeclInfo(VariableDeclInfo variableDeclInfo) {
   @Override
   public void visitParameterDecl(ParameterDecl parameterDecl) {
     visitChildFromParent(parameterDecl.getType(), parameterDecl);
-    visitChildFromParent(this::visitArrayInfo, parameterDecl.getArrayInfo(), parameterDecl);
```
paulthomson: Could there potentially be a new test for this? I may not quite understand what is happening.

common/src/main/java/com/graphicsfuzz/common/util/CannedRandom.java
afd: Yes, I did; apologies for not mentioning it in the commit message.

ast/src/main/java/com/graphicsfuzz/common/ast/visitors/StandardVisitor.java
afd: Done.

Issue #165, opened 2018-12-11
Document WebUI image borders.


afd:
The image borders in the WebUI indicate whether a result is likely a bug (red) or probably just round-off error (blue).

We should make sure the documents properly explain this, to avoid situations where folks run a reduction on a blue-border image.

PR #302, opened 2019-02-27
Document colors of result tables with a legend in the UI

hevrard:
The legend only shows up in pages which have a result table.

NOTE: this change revealed that some colors were too close to each other, so I also updated the colors.

The legend looks like this:
legend


hevrard:
Fix #165

PR #308, opened 2019-02-27
Fixed --skip-render in runspv and add more compute shader tests

afd:


python/src/main/python/cmd_helpers.py
```diff
@@ -33,20 +33,37 @@ def try_get_jar_bin_dirs(install_root):
             return os.path.abspath(bin_dir), os.path.abspath(jar_dir)
         return None
 
-    # Perhaps we are running from the zip.
-    res = try_get_jar_bin_dirs(path(os.path.dirname(HERE), os.pardir))
+    # Perhaps we are running from the IDE.  Check this first, since the deployed files are likely also present if
+    #     # running from the IDE.
```
paulthomson: nit: double comment

afd: Thanks - fixed.  Please re-approve.

PR #316, opened 2019-02-27
runspv: improve subprocess_helper

### LINKED ISSUE #301: runspv/glsl-to-spv-worker: glslangValidator, spirv-opt, etc. can fail. Handle this somewhere.
Right now I can see this in the worker trace:

```
No job
#### Image job: reference
test.vert
reference.frag
ERROR: reference.frag:3: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.


ERROR: Linking fragment stage: Missing entry point: Each stage requires one entry point

SPIR-V is not generated for failed compile or link
Send back, results status: 70
No job
```

Yet the log passed on to the server is:
```
Start: reference
ERROR
COMMAND: /usr/local/google/home/hevrard/work/graphicsfuzz/graphicsfuzz/target/graphicsfuzz/python/drivers/../../bin/Linux/glslangValidator reference.frag -V -o reference.frag.spv
RETURNCODE: 2
STDOUT: None
STDERR: None
```

### LINKED ISSUE #306: vkrun: amber desktop error messages are not captured
This is probably similar to #301

paulthomson:
Probably fix #306 (amber desktop error messages are not captured)

Probably fix #301

stdout and stderr are now captured by default (but not printed, as this is too verbose). We print the command and returncode, which is helpful for understanding what runspv is doing. "check=True" by default.

### DIFF
--- a/python/src/main/python/drivers/runspv.py
+++ b/python/src/main/python/drivers/runspv.py
@@ -185,6 +185,14 @@ def spirvdis_path():
     return tool_on_path('spirv-dis')
 
 
+def adb_path():
+    if 'ANDROID_HOME' in os.environ:
+        adb = os.path.join(os.environ['ANDROID_HOME'], 'platform-tools', 'adb')
+        if os.path.isfile(adb):
+            return adb
+    tool_on_path('adb')
+
+
 def remove_end(str_in: str, str_end: str):
     assert str_in.endswith(str_end), 'Expected {} to end with {}'.format(str_in, str_end)
     return str_in[:-len(str_end)]
@@ -263,7 +271,7 @@ def adb_helper(
     stdout: Union[None, int, IO[Any]]=subprocess.PIPE,
     verbose=False
 ):
-    adb_cmd = ['adb'] + adb_args
+    adb_cmd = [adb_path()] + adb_args
 
     try:
         return subprocess_helper(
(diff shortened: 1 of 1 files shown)


python/src/main/python/drivers/runspv.py
```diff
@@ -220,28 +257,26 @@ def prepare_shader(output_dir: str, shader: Optional[str]):
 TIMEOUT_APP = 30
 
 
-def adb_helper(adb_args: List[str], check, stdout: Union[None, int, IO[Any]]):
+def adb_helper(adb_args: List[str], check, stdout: Union[None, int, IO[Any]]=subprocess.PIPE):
```
hevrard: Maybe add a verbose argument here that can be passed to subprocess_helper() ?

paulthomson: Sounds good. Done. We may also want to actually make some calls verbose, but it seems to look OK from some quick testing.

python/src/main/python/drivers/runspv.py
```diff
@@ -70,25 +70,62 @@ def print(s):
     orig_print(s, flush=True)
 
 
-def subprocess_helper(cmd: List[str], check=False, timeout=None, stdout=None):
+def convert_and_print_stdout_stderr_returncode(result, verbose=False):
```
afd: Let's try to use argument types throughout the Python worker code as much as possible.

paulthomson: I agree. I actually avoided it here because "result" is a mixture, but I can use Union

PR #322, opened 2019-02-28
runspv: stdout/err to logfile; don't write status twice

### LINKED ISSUE #319: runspv: writes status twice for Android

### LINKED ISSUE #320: runspv: Android: does not capture amber stdout and stderr
We did not need to do this previously because we were just launching an apk. But now we should.

paulthomson:
Fix #320 
Fix #319


paulthomson:
More changes incoming

PR #325, opened 2019-02-28
Exploit passthrough option in Amber to avoid passing a vertex shader

afd:


python/src/main/python/drivers/runspv.py
```diff
@@ -704,8 +704,11 @@ def amberscriptify_image(vert, frag, uniform_json):
     script += '[require]\n'
     script += 'fbsize 256 256\n'
 
-    script += '[vertex shader spirv]\n'
-    script += spv_get_disassembly(vert)
+    if vert:
```
paulthomson: I can't figure out if the convention is "is not None". LGMT though.

PR #328, opened 2019-02-28
Add metrics summary data to WebUI.

### LINKED ISSUE #321: When viewing a result, information about the image metrics should be available
When you view an image result, you should be presented with info about the image metrics so that if the metrics disagree you can see which one thinks the images are different and decide to use that one for a reduction if appropriate.

Right now it's only possible to get this information via a back door.

I'm thinking something like this mock-up (though I am unsure whether we are currently looking at the results of PSNR):

webui

paulthomson:
Fix #321 

image

### DIFF
--- a/server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
+++ b/server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
@@ -882,9 +882,12 @@ public class WebUi extends HttpServlet {
 
     String pngPath = variantDir + variant + ".png";
     File pngFile = new File(pngPath);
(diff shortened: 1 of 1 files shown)


paulthomson:
OK I have made some changes. The metrics will only show if the status was success and there is a difference. The histogram distance is shown. I hate the code, but in the end, we are accessing untyped JSON; I would rather invest in making this typed than trying to build a nice layer of abstraction. Even though I know I have just done a layer of abstraction somewhat! The layer here was just a necessity so that I can reuse the logic that _accesses_ the JSON in three places, because it is already unnecessarily brittle.

PR #333, opened 2019-02-28
Use fence timeout of 60s with Amber.  Also adapted construction of c…

afd:
Instruct Amber to use a generous fence timeout for image and compute jobs.


paulthomson:
Looks good. Check the commit message before merging: should be 60s.

PR #340, opened 2019-03-01
Move shader preparation (including spirv-opt) into runspv, and added spirv-opt tests

afd:


python/src/main/python/drivers/runspv.py
```diff
@@ -45,6 +46,9 @@
 
 SKIP_RENDER_OPTION_HELP = 'Compile shaders but do not actually run them.'
 
+SPIRV_OPT_OPTION_HELP = 'Enable spirv-opt with these optimization flags.  Multiple arguments ' \
```
paulthomson: nit: Use parens to avoid needed slash

afd: Done.

Issue #338, opened 2019-02-28
Linux amber does not get properly killed


hevrard:
We observed several amber commands still running (and eating all memory), although only one should be running at a time. I suspect the python subprocess timeout does not properly kill the amber process in case of timeout.

We should double check that a real SIGKILL is sent to amber when it times out.

afd:
@hevrard or @paulthomson , can one of you take this?

afd:
Tentatively assigned @hevrard

hevrard:
Thanks @paulthomson for the joined investigation!

It looks like python subprocess timeout does not mix well with catchsegv which spawns new processes, such that when python kills the subprocess due to a timeout, the top catchsegv process gets killed but its children (another catchsegv and amber) happily continue to run.

We can replace catchsegv by `gdb -batch -ex run -ex backtrace --args amber ...`, this will produce a stack trace when a segfault occurs; thanks Paul for this one-liner!

Now on a small experiment I'm afraid gdb adds a significant delay (+4 sec per run) probably due to loading symbols from amber, I'll try to see if we can bypass this issue.

hevrard:
Running amber release/debug builds, with/without under gdb, on a single non-trivial shader leads to these results:

amber-release: 0.3 sec
amber-debug: 0.7 sec
gdb amber-release: 1.3 sec
gdb amber-debug: 4.9 sec

Using gdb leads to a significant slowdown, 1 second in the best case.

I suggest we use core dump and retrieve the backtrace afterwards only in case of crash.

I'll wait for the PR tweaking runspv subprocess() usage to be merged before working on a fix.

hevrard:
The gdb wrapper leads to a weird problem where, *after* a test timeouts, then the next test will be suspended as if it received SIGSTOP (equivalent to Ctrl-Z in a terminal):

```
#### Image job: variant_100
Exec:['/gpu/graphicsfuzz/graphicsfuzz/target/graphicsfuzz/python/drivers/../../bin/Linux/glslangValidator', '-V', 'variant_100.frag', '-o', '/gpu/release-1-1/host-amber/variant_100.frag.spv']
RETURNCODE: 0
Exec:['/gpu/graphicsfuzz/graphicsfuzz/target/graphicsfuzz/python/drivers/../../bin/Linux/spirv-dis', '/gpu/release-1-1/host-amber/variant_100.frag.spv']
RETURNCODE: 0
[...]

PR #343, opened 2019-03-01
Remove usage of catchsegv as it is not killed properly in case of timeout

hevrard:
Do not try to get the stack trace on Linux. We will revisit this later, see #338 for the discussion.

### DIFF
--- a/python/src/main/python/drivers/runspv.py
+++ b/python/src/main/python/drivers/runspv.py
@@ -151,7 +151,7 @@ def subprocess_helper(
         )
     except subprocess.TimeoutExpired as ex:
         convert_stdout_stderr(ex)
-        # No returncode in case of timeout.
+        # no returncode to log in case of timeout
         log_stdout_stderr(ex)
         raise ex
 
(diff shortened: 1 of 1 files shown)


afd:
I will look, but would prefer @paulthomson to do an approving review.

python/src/main/python/drivers/runspv.py
```diff
@@ -231,13 +237,24 @@ def spirvopt_path():
...
-        # Didn't find catchsegv on path; that's OK
-        pass
+# FIXME: using gdb wrapping leads to a weird issue where, once a test times out,
+# subsequent tests leads to the whole worker being SIGSTOP'ed.
+# See https://github.com/google/graphicsfuzz/issues/338
+
+# def maybe_add_gdb(cmd: List[str]) -> None:
+#     # If gdb is available, use it to obtain backtraces in case of segfault
+
+#     # TODO: using gdb as a wrapper adds at least a second-long overhead to each
```
afd: Can you change TODO to TODO(issue number) where 'issue number' is the number of the associated issue for this optimization idea?

afd: (I like the reference to the issue above for the FIXME.  But this sound like it deserves a different issue.)

paulthomson: I propose we make this PR:

* Remove use of `catchsegv`
* Fix issue with subprocess_helper timeout

And we don't commit the commented out code; we instead create an issue, and put the code there or leave the code in a branch.

hevrard: See #349 for the fix of timeout.

paulthomson: Looks good. Let's make this just:

* Remove use of `catchsegv`

PR #346, opened 2019-03-03
Compute shader scripts

afd:
Scripts to support analysis of compute shader family results.


paulthomson:
Reading it: looks good.

However, the tests fail for me. There are also some `return None`, `return 1`, etc. I would much prefer that we raise exceptions at the location of the error, and only convert to error codes at the very top level of command line tools: `if __name__ == '__main__':`. Thus, the tests will raise exceptions and I can get a stack trace.

paulthomson:
Oh I think there are also a few PEP warnings; one about a duplicate test might be serious.

afd:
Thanks for the feedback.  I have adapted things so that exceptions are raised in all cases where inputs are invalid.

For the diffing: I don't want to raise an exception if the inputs are valid but different.  That was where I was returning 0 or 1.  I have changed this so that the diff functions return a boolean and a message.

I'm surprised the tests failed for you - would be good to know which ones.

python/src/main/python/interestingness/compute_interesting_diff.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
paulthomson: Was this file supposed to be committed?

afd: No, it wasn't - good catch.  It's for a future PR.  I'll do a push where I delete it, agreed?

paulthomson:
Ah I see. Apologies; I think misunderstood the code and may have even misinterpreted the results due to the stderr output. On the other hand, I like the changes! And it was good to fix those duplicated/misnamed tests. LGTM, except for that empty file.

python/src/main/python/interestingness/compute_interesting_diff.py
paulthomson: Sounds good!

afd:
OK, that file has been removed and I have made one small pretty-printing change.

PR #348, opened 2019-03-03
Reworked reducer

afd:
This replaces the reducer with a re-worked version that more closely resembles delta debugging-style algorithms.  A number of experiments suggest that it typically performs on par with the old reducer, and has been shown to be significantly faster on some long-running reductions.  The form of this new reducer is also likely easier to experiment with for future optimization.

In this PR the old reducer is still there and can be enabled via source code changes (not a dynamic option).  I suggest we keep it around for a while, so that it is kept up-to-date with other changes in the code base, until we are sure we no longer need it.


paulthomson:
Awesome! Looks like there are a few bug fixes in this PR. I don't think we generally care about separating fixes into separate PRs, especially as they might be dependent, but perhaps we should try to add issues for these, and mark them all as fixed by this PR; it might be helpful later when revisiting the related issues.

afd:
Agreed that in principle it would be good to file and close issues for the other problems this PR fixed.  That said, this was a bit of an exceptional PR due to the time lag.

PR #351, opened 2019-03-04
Re-enable amber stack trace on Linux with proper timeout killing

### LINKED ISSUE #350: Enable catchsegv on Linux with proper group process handling
We should be able to use catchsegv with proper process group handling for the kill signal.

hevrard:
Fix #350

### DIFF
--- a/python/src/main/python/drivers/runspv.py
+++ b/python/src/main/python/drivers/runspv.py
@@ -94,7 +94,10 @@ def log(message: str) -> None:
 
 def log_stdout_stderr(
     result: Union[
-        subprocess.CalledProcessError, subprocess.CompletedProcess, subprocess.TimeoutExpired],
+        subprocess.CalledProcessError,
+        subprocess.CompletedProcess,
+        subprocess.TimeoutExpired,
+    ],
 ) -> None:
 
     log('STDOUT:')
@@ -194,7 +197,7 @@ def run_catchsegv(cmd: List[str], timeout=None, verbose=False) -> str:
             status = 'CRASH'
         log_returncode(proc)
 
-        if verbose:
+        if verbose or proc.returncode != 0:
             if outs is not None:
                 outs = outs.decode(encoding='utf-8', errors='ignore')
             if errs is not None:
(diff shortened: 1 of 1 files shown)


python/src/main/python/drivers/runspv.py
```diff
@@ -921,12 +972,16 @@ def run_image_amber(
...
         cmd.append(amberscript_file)
         status = 'SUCCESS'
-        try:
-            subprocess_helper(cmd, timeout=TIMEOUT_RUN, verbose=True)
-        except subprocess.TimeoutExpired:
-            status = 'TIMEOUT'
-        except subprocess.CalledProcessError:
-            status = 'CRASH'
+
+        if 'catchsegv' in cmd[0]:
```
paulthomson: I would prefer making `maybe_add_catchsegv` return a bool and then checking this here

python/src/main/python/drivers/runspv.py
```diff
@@ -1170,6 +1225,7 @@ def run_compute_amber(
 
     else:
         cmd = []
+        maybe_add_catchsegv(cmd)
```
paulthomson: Same here

python/src/main/python/drivers/runspv.py
```diff
@@ -1181,12 +1237,16 @@ def run_compute_amber(
...
 
         status = 'SUCCESS'
-        try:
-            subprocess_helper(cmd, timeout=TIMEOUT_RUN, verbose=True)
-        except subprocess.TimeoutExpired:
-            status = 'TIMEOUT'
-        except subprocess.CalledProcessError:
-            status = 'CRASH'
+
+        if 'catchsegv' in cmd[0]:
```
paulthomson: Same as above.

python/src/main/python/drivers/runspv.py
paulthomson: Fixed

python/src/main/python/drivers/runspv.py
paulthomson: Fixed

python/src/main/python/drivers/runspv.py
```diff
@@ -170,6 +173,47 @@ def subprocess_helper(
...
+    proc = subprocess.Popen(
+        cmd,
+        start_new_session=True,  # this creates a process group for all children
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+
+    try:
+        outs, errs = proc.communicate(timeout=TIMEOUT_RUN)
+        if proc.returncode != 0:
```
paulthomson: Should log stdout and stderr if non zero returncode

python/src/main/python/drivers/runspv.py
```diff
@@ -170,6 +173,47 @@ def subprocess_helper(
...
+            log(outs)
+            log('')
+            log('STDERR:')
+            log(errs)
+            log('')
+
+    except subprocess.TimeoutExpired as ex:
+        # Kill the whole process group to include all children of catchsegv
+        os.killpg(proc.pid, signal.SIGKILL)
+        status = 'TIMEOUT'
```
paulthomson: Should probably log stdout and stderr (as much as possible) on timeout

python/src/main/python/drivers/runspv.py
paulthomson: Fixed

python/src/main/python/drivers/runspv.py
paulthomson: Fixed

PR #354, opened 2019-03-04
glsl-to-spv-worker: use output_dir and other fixes

### LINKED ISSUE #301: runspv/glsl-to-spv-worker: glslangValidator, spirv-opt, etc. can fail. Handle this somewhere.

### LINKED ISSUE #317: runspv (android) with glsl-to-spv-worker: pulls files into the current directory, which may clash with previously pulled files
It looks like we try to delete old files first, but we might miss some and the legacy worker in particular pulls the contents of an entire directory. We should: delete "graphicsfuzz/", pull "graphicsfuzz/", refer to result files inside "graphicsfuzz/". We will need to update the test scripts too.

### LINKED ISSUE #352: runspv: legacy host: skip-render flag is not passed

### LINKED ISSUE #353: runspv and glsl-to-spv-worker: some files are still written to current directory instead of output dir

### LINKED ISSUE #355: runspv: use `tool_on_path` for all binaries run from path
Doing e.g. `subprocess.run(['vkworker', ...], ...)` (with vkrunner on your PATH) will often work on Linux, but it is generally unreliable (e.g. modifying `PATH` from within the Python script might not work, even though it should) and supposedly does not work cross-platform. We should use `tool_on_path` everywhere.

paulthomson:
Fix glsl-to-spv-worker and runspv to consistently use the output_dir parameter, instead of (sometimes) writing files to the current directory.
runspv: remove more accidental direct writes to "logfile", instead of using `log('...')`. 
Base output directory on job name.
Clear output directory before start; avoids name clashes from previous results; avoids having to delete individual files that could be missed.
Do not use os.chdir.
Fix skip_render flag for legacy worker on host platform. 
Make `dump_info_host_legacy` just raise an exception if it fails instead of writing to STATUS file; we will catch the exception anyway. 
glsl-to-spv-worker: use `open_helper`. 

Fix #317 files may clash with previously pulled files. 
Fix #352 skip render flag not passed on legacy host. 
Fix #353 some files are still written to current directory instead of output dir. 
Fix #301 glslangValidator, spirv-opt, etc. can fail. Handle this somewhere. 
Fix #355 use `tool_on_path` for all binaries run from path.

### DIFF
--- a/python/src/main/python/drivers/glsl-to-spv-worker.py
+++ b/python/src/main/python/drivers/glsl-to-spv-worker.py
@@ -62,7 +62,7 @@ def print(s):
 
 
 def write_to_file(content, filename):
-    with open(filename, 'w') as f:
+    with runspv.open_helper(filename, 'w') as f:
         f.write(content)
 
 
@@ -144,7 +144,7 @@ def do_image_job(
 
     # Set runspv logger. Use try-finally to clean up.
 
-    with open(log_file, 'w', encoding='utf-8', errors='ignore') as f:
+    with runspv.open_helper(log_file, 'w') as f:
         try:
             runspv.log_to_file = f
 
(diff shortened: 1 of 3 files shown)


paulthomson:
I marked some failing tests as skipped; they are transparency issues, plus the spirv-opt test fails for me; possibly a bug in my device's drivers, as the image do indeed look slightly different.

PR #362, opened 2019-03-05
Adapted mandelbrot shaders to make them more colourful and more blocky.

afd:


paulthomson:
Could update test shaders too?

PR #365, opened 2019-03-05
WIP Handle compute shader results in UI by pointing to documentation

### LINKED ISSUE #347: Remove compute shaders from results view in web UI
It is handy to be able to launch compute shader families using the web UI.  But currently their results are then available for viewing, but are not presented in a meaningful way.  Let's make it so that their results are not shown at all.  We can decide in future if we want them to be shown and implement proper support at that time.

hevrard:
Fix #347 
WIP as we have to update the URL to point to compute shader documentation.


afd:
@paulthomson can you review this please?

PR #369, opened 2019-03-05
Compute shader samples and documentation

afd:
Added additional compute shader samples, documented how to work with compute shaders, and added some example interestingness tests.


docs/glsl-fuzz-walkthrough.md
```diff
@@ -570,6 +594,8 @@ was enough to cause the wrong image to be rendered.
 
 ## Queuing a no image reduction
 
+> Reductions for the case where no SSBO is produced by a compute shader cannot be launched via the Web UI.  [See how to launch them via the command line](#Performing-a-"no-image"-reduction-on-a-fragment-or-compute-shader).
```
hevrard: Are the double quotes (") OK in an internal link? spotted in `(#Performing-a-"no-image"-reduction-on-a-fragment-or-compute-shader)`

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+
+This corresponds to a value of 88 for the `result` field of the SSBO, and values of 28, 12, 14, etc., for the elements of the `data` array field.
+
+### Showing the SSBO for a compute shader result
+
+To view the SSBO for a compute shader result in pretty-printed form, do:
+
+```sh
+# The 'show' argument specifies that we wish to view the SSBO for the given result file
+inspect_compute_results show /path/to/results/reference_or_variant.info.json
```
hevrard: In the first release we made all command names use hyphens rather than underscore (e.g. `glsl-to-spv-worker`), I suggest we do the same for `inspect_compute_results` -> `inspect-compute-results`

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+```
+
+### Comparing SSBOs for two compute shader results
+
+To compare SSBOs for two result files from the same shader family (typically the reference and a variant), do:
+
+```sh
+# The 'exactdiff' argument specifies that we wish to compare the SSBOs for the given result files, and that they should be identical
+inspect_compute_results exactdiff /path/to/result/reference.info.json /path/to/result/variant_XXX.info.json
```
hevrard: hyphenize?

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+```
+
+This will exit with code 0 if and only if the SSBOs associated with the two result files match *exactly*.  Otherwise, a mismatch will be reported and a non-zero exit code returned.
+
+To tolerate some degree of floating-point difference when comparing SSBO entries, do:
+
+```sh
+# The 'fuzzydiff' argument specifies that we wish to compare the SSBOs for the given result files, and that we want to allow some tolerance of numerical differences
+inspect_compute_results fuzzydiff /path/to/result/reference.info.json /path/to/result/variant_XXX.info.json
```
hevrard: hyphenize?

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+```
+
+Fuzzy diffing uses the Python `isclose` function to compare numbers.  You can pass `--rel_tol` and `--abs_tol` to `inspect_compute_results` to control the relative and absolute tolerance used by `isclose`; see the [Python math documentation](https://docs.python.org/3/library/math.html) for details of the `rel_tol` and `abs_tol` arguments to `isclose`.
+
+```sh
+# Request a fuzzy diff with relative tolerance of 1e-5 and absolute tolerance of 0.1
+inspect_compute_results fuzzydiff --rel_tol=1e-5 --abs_tol=0.1 /path/to/result/reference.info.json /path/to/result/variant_XXX.info.json
```
hevrard: hyphenize?

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+```
+
+To see the default values for `--rel_tol` and `--abs_tol`, do:
+
+```sh
+inspect_compute_results --help
```
hevrard: hyphenize?

docs/glsl-fuzz-walkthrough.md
```diff
@@ -698,6 +724,101 @@ the reduction of this variant leads to the following files:
...
+```
+
+## Viewing summary results for a compute shader family
+
+To see a summary of the results for a compute shader family, do e.g.:
+
+```sh
+report_compute_shader_family_results family_vulkan_compute_frag2comp-0005-squares
```
hevrard: hyphenize?

docs/glsl-fuzz-walkthrough.md
```diff
@@ -858,11 +981,37 @@ Explanation:
...
+# Copy the .info.json file for the reference into your working directory
+cp /path/to/reference.info.json .
+# Make a copy of the template interestingness test for compute shader differences
+cp graphicsfuzz/python/interestingness/compute_interesting_diff.py .
+# Make the script executable
+chmod +x compute_interesting_diff.py
+
+# Edit 'REFERENCE_RESULTS', 'WORKER_NAME' and 'SERVER_URL' in 'compute_interesting_diff.py' to suit your needs
+
+# Run a reduction using 'compute_interesting_diff.py' to decide whether each reduced shader is interesting.  Results will be placed in 'someoutputdir'.  --preserve-semantics ensures that only semantics-preserving changes are made during reduction.
```
hevrard: This is a very long line shell comment line in a code section in markdown, which will not be auto-wrapped, so I suggest to wrap it to approx. 80 char width.

docs/glsl-fuzz-walkthrough.md
```diff
@@ -858,11 +981,37 @@ Explanation:
...
+
+### Performing a bad result reduction for a compute shader
+
+If you have a variant compute shader that produces a different SSBO result compared with its associated reference, you can reduce the problem as follows.
+
+```sh
+# Copy the .info.json file for the reference into your working directory
+cp /path/to/reference.info.json .
+# Make a copy of the template interestingness test for compute shader differences
+cp graphicsfuzz/python/interestingness/compute_interesting_diff.py .
```
paulthomson: The chmod should not be necessary. I believe the file is made executable before being added to the zip. The instructions here are unlikely to work on Windows even if they use a bash shell because the reducer cannot launch `./compute_interesting_diff.py` on Windows; we should include bash and cmd wrappers and copy them too. 

So I suggest the following:

* Remove the chmod command from the docs.
* Make the python scripts executable before committing; it is not necessary, but it is good to be consistent. 
* Add bash and cmd wrappers for the scripts in `interestingness`. 
* Docs: change the `cp` to include the appropriate wrappers. 
* Docs: change the glsl-reduce command to remove `.py`; use `./compute_interesting_diff`. 
* Same for `compute_interesting_same.py`.

docs/glsl-fuzz-walkthrough.md
afd: VS code was OK with this, but it seems that it doesn't work on GitHub.  I've removed the quotes.

docs/glsl-fuzz-walkthrough.md
afd: I agree, and done (though a bit tricky since another Python file needs to import inspect_compute_results and a hyphen cannot appear in an import).

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

docs/glsl-fuzz-walkthrough.md
afd: Done.

PR #380, opened 2019-03-06
Add initial version of libFuzzer custom mutator server

jonathanmetzman:
This adds a working server that can accept shaders over TCP from a libFuzzer custom mutator and returns a mutated shader.

I've confirmed this works by modifying [ANGLE's translator fuzzer](https://cs.chromium.org/chromium/src/testing/libfuzzer/fuzzers/BUILD.gn?type=cs&q=angle_translator_fuzzer&sq=package:chromium&g=0&l=340) to use this server instead of using libFuzzer's standard mutations.
I still need to figure out how those changes will be upstreamed to ANGLE (I want to do it in a way that doesn't mess things up on ClusterFuzz).

This implementation may be replaced with something faster (e.g. shared memory).

Other details of the implementation will probably change, these include: 
* Currently the server listens on port `8666` and can only accept one connection. This doesn't allow parallel use of libFuzzer and other libFuzzer features.
* Whether we even try to support libFuzzer's custom crossover
* How we handle sizes on 32 bit platforms. Currently we treat sizes as 64 bits (`long`) because we use `size_t` (64 bits on x86_64 but 32 on x86) on the libFuzzer side.

### DIFF
--- a/docs/libFuzzer-integration.md
+++ b/docs/libFuzzer-integration.md
@@ -140,12 +140,14 @@ extern "C" size_t LLVMFuzzerCustomMutator(uint8_t* data, size_t size,
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   std::string shader(reinterpret_cast<const char*>(data), size);
+  // Give the user some feedback since the coverage won't grow much in an empty
+  // fuzzer.
   std::cout << shader << std::endl;
   return 0;
 }
 EOF
 
-# Must use clang-6.0 or greater (sudo apt install clang-6.0 && clang-6.0 ...).
+# Must use clang-6.0 or greater (sudo apt install clang-6.0 && clang-6.0 -fsanitize=fuzzer...).
 clang++ -g -fsanitize=fuzzer,address fuzzer.cc -o fuzzer
 ```
 
--- a/docs/libFuzzer-integration.md
+++ b/docs/libFuzzer-integration.md
@@ -7,6 +7,10 @@ GraphicsFuzz can be used with [libFuzzer](http://llvm.org/docs/LibFuzzer.html)
 as a [custom mutator](https://cs.chromium.org/chromium/src/third_party/libFuzzer/src/FuzzerInterface.h)
 for fuzzing that is both coverage-guided and [structure-aware](https://github.com/google/fuzzer-test-suite/blob/master/tutorial/structure-aware-fuzzing.md).
 
+In examples below, the `graphicsfuzz/` directory is the unzipped release. If
+building from source, this directory can be found at
+`graphicsfuzz/target/graphicsfuzz/`.
+
 ## Using the Integration
 
 1. Build GraphicsFuzz.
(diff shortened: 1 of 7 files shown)


jonathanmetzman:
cc @kcc who likes hearing about custom mutator developments.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+/**
+ * Server that can accept shaders from a libFuzzer custom mutator and send back a mutated shader.
+ */
+public class CustomMutatorSever {
+  public static void main(String[] args) {
+    try {
+      // TODO(metzman): If we don't switch from TCP, allow this to be configurable.
```
afd: Can you change TODO(metzman) to TODO(N) where N is the id of an issue in the GraphicsFuzz issue tracker?

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+      runServer(8666);
+    } catch (IOException | ParseTimeoutException | InterruptedException exception) {
+      exception.printStackTrace();
+      System.exit(1);
+    }
+  }
+
+  private static void runServer(int port)
+      throws IOException, ParseTimeoutException, InterruptedException {
+    byte[] headerBuff = new byte[28];
```
afd: Please use final in declarations wherever possible.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+    ServerSocket serverSocket = new ServerSocket(port);
+    Socket socket = serverSocket.accept();
+    InputStream inputStream = socket.getInputStream();
+    OutputStream outputStream = socket.getOutputStream();
+    while (true) {
+      // TODO(metzman): Figure out a better way to handle waiting for the header to arrive.
```
afd: Similar comment to the above for TODO(metzman).

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+      // int before creating the array. This is probably a non-issue because libFuzzer is unlikely
+      // to ever give us a shader larger than Integer.MAX_VALUE (e.g. ~2G).
+      assert size1 <= Integer.MAX_VALUE && size2 <= Integer.MAX_VALUE;
+
+      // GraphicsFuzz can't do anything with this unfortunately.
+      long maxOutSize = headerByteBuffer.getLong();
+
+      int libfuzzerSeed = headerByteBuffer.getInt();
+
+      // Read the Shader.
```
afd: Shader -> shader

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+      inputStream.read(headerBuff, 0, headerBuff.length);
+      ByteBuffer headerByteBuffer = ByteBuffer.wrap(headerBuff);
+      headerByteBuffer.order(ByteOrder.LITTLE_ENDIAN);
+
+      long size1 = headerByteBuffer.getLong();
+      // Support two sizes for now so that one message format can be used for both CustomMutate and
+      // CustomCrossOver.
+      long size2 = headerByteBuffer.getLong();
+      assert size2 == 0;
+
```
afd: Please expand the above comments to say a bit more about what the two sizes mean.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+      int libfuzzerSeed = headerByteBuffer.getInt();
+
+      // Read the Shader.
+      byte[] inputShaderBuff = new byte[((int) size1) + ((int) size2)];
+      socket.getInputStream().read(inputShaderBuff, 0, inputShaderBuff.length);
+      String inputShader = new String(inputShaderBuff);
+
+      try {
+        // Now mutate the shader.
+        TranslationUnit tu = ParseHelper.parse(inputShader, ShaderKind.FRAGMENT);
```
afd: Please document the assumption that the shader is a fragment shader.  I guess from a libFuzzer perspective the shader kind is unimportant?  (And indeed we then usually use ShaderKind.FRAGMENT as a default.)

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
...
+        Mutate.mutate(tu, new RandomWrapper(libfuzzerSeed));
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        try (PrintStream stream = new PrintStream(byteArrayOutputStream, true, "UTF-8")) {
+          PrettyPrinterVisitor.emitShader(
+              tu,
+              Optional.empty(),
+              stream,
+              PrettyPrinterVisitor.DEFAULT_INDENTATION_WIDTH,
```
afd: It occurs to me that if minimzing shader size matters, an indentation width of zero might be a good thing.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2018 The GraphicsFuzz Project Authors
```
afd: @paulthomson do you have thoughts about whether generator/src/main/java/com/graphicsfuzz/generator/tool/ is a suitable location for this class?  Would it perhaps be better for us to add a new Maven module for libFuzzer integration?

jonathanmetzman:
> Thanks very much for this!
> 
> Beyond the comments on the code, would you be able to contribute some simple documentation explaining what one needs to do to try this out? I would be keen to try it out as part of reviewing it. You could put it under docs. E.g. libFuzzer-integration.md.

Sure. I definitely should have done this initially, since it's probably very unclear what any of this stuff means right now.
I'll do this with my next uploads.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
jonathanmetzman: I'm not sure I understand but I'm not sure eliminating whitespace is a helpful minimization here.
I think minimal testcases are most useful because the odds that random, totally arbitrary mutations result in new behavior are better when the testcase is reduced. But in structure aware fuzzing this isn't always important since the fuzzer knows not to mutate whitespace. What do you think?

This comment also makes me realize that single testcase minimization won't work well since the best we can do is hope mutations reduce the size of shaders, but this issue can be fixed later.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
jonathanmetzman: The two sizes was a bad decision since it wasn't needed yet (may not be ever). So I got rid of the second one and made it clearer what the first one is.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
jonathanmetzman: Done

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
jonathanmetzman: It does sort of matter for libFuzzer in this particular case. I changed things so libFuzzer can instruct GraphicsFuzz to use a fragment or vertex.

docs/libFuzzer-integration.md
```diff
@@ -0,0 +1,179 @@
...
+```bash
+java -ea -cp graphicsfuzz/target/graphicsfuzz/jar/tool-1.0.jar com.graphicsfuzz.generator.tool.CustomMutatorServer
+Listening on port: 8666
+```
+
+3. Build a fuzz target (in a new shell, since the previous example won't
+   terminate). An example is provided below:
+
+```bash
+cat << EOF > fuzzer.cc
```
jonathanmetzman: Maybe this code should live in its own file and I should instruct people to download it?
What do you think?

docs/libFuzzer-integration.md
```diff
@@ -0,0 +1,184 @@
...
+  std::string shader(reinterpret_cast<const char*>(data), size);
+  // Give the user some feedback since the coverage won't grow much in an empty
+  // fuzzer.
+  std::cout << shader << std::endl;
+  return 0;
+}
+EOF
+
+# Must use clang-6.0 or greater (sudo apt install clang-6.0 && clang-6.0 -fsanitize=fuzzer...).
+clang++ -g -fsanitize=fuzzer,address fuzzer.cc -o fuzzer
```
jonathanmetzman: By the way, this command fails on my machine, I need to use `clang-6.0` instead of clang++, but I think the clang on our boxes is older than most.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
afd: I just meant that if the overhead of transferring shaders between libfuzzer and graphicsfuzz is high, then minimizing size of the shader text could help a bit, and minimizing whitespace might help with that.  A shader with deeply-nested conditionals can end up having a lot of whitespace in its text representation due to indentation.  But perhaps this is a non-issue in practice.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
paulthomson: I feel it is probably OK here at this stage.

jonathanmetzman:
I've done a bit of profiling and measurement it seems like the interprocess communication is not the main overhead during mutation. I could be wrong since profiling two processes was a bit hard.
I determined this by changing the server to simply read the shader from libFuzzer and write back 0 as the response. Under this setup the only I/O calls that we lose are the server sending the shader and the client reading the shader. This setup gives us 75,000 execs/sec instead of ~10, leading me to think that GraphicsFuzz is taking most of the time.

jonathanmetzman:
Actually ignore my previous comment about profiling, my code was incorrect.
Now that I have fixed it the communication seems to be the bottleneck. If the client sends a shader to the server and reads an empty shader back (ie: we run the mutation server but always send empty shaders back), we end up with 20 execs/sec instead of 10/sec.
So it seems like real speed up will come from making communication faster.

generator/src/main/java/com/graphicsfuzz/generator/tool/CustomMutatorSever.java
jonathanmetzman: OK Done

jonathanmetzman:
I've added some possibly incomplete instructions on using JNI for libFuzzer to talk Java.
Feel free to ignore them and just do TCP as was planned.

JNI sped up fuzzing by quite a bit, but things are still too slow. However, I think not doing any preprocessing (I think that can be done in another PR) gets execution speed to a good place.
I think there's value in having both a TCP and a JNI implementation. But given the performance benefits I think the JNI implementation should probably be preferred.
We can discuss more tomorrow.

docs/libFuzzer-integration.md
jonathanmetzman: Done.

docs/libFuzzer-integration.md
paulthomson: Note for us (I doubt we need to add this to the docs): As well as installing `clang-6.0`, install `libstdc++-8-dev`.

jonathanmetzman:
> LGTM! I am happy with the file locations. The .cc files need license headers to pass the CI. See:
> `vulkan-worker/src/common/vulkan_worker.cc` for a header you can copy, although change the year to 2019

Done.

PR #402, opened 2019-03-11
Refined the reducer's facilities for unwrapping block statements.  It…

afd:
… did not previously unwrap a statement that contained declaration statements.  Now it does, so long as doing so will not introduce name clashes.  This had some small knock-on effects on the reduction of live color writes, which can now get unwrapped so that they appear in the outer-most block of a function.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunities.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunities.java
@@ -17,6 +17,7 @@
 package com.graphicsfuzz.reducer.reductionopportunities;
 
 import com.graphicsfuzz.common.ast.TranslationUnit;
+import com.graphicsfuzz.common.ast.decl.VariableDeclInfo;
 import com.graphicsfuzz.common.ast.stmt.BlockStmt;
 import com.graphicsfuzz.common.ast.stmt.DeclarationStmt;
 import com.graphicsfuzz.common.ast.stmt.DoStmt;
@@ -27,7 +28,11 @@ import com.graphicsfuzz.common.ast.stmt.Stmt;
 import com.graphicsfuzz.common.transformreduce.ShaderJob;
 import com.graphicsfuzz.common.util.ListConcat;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class UnwrapReductionOpportunities
       extends ReductionOpportunitiesBase<UnwrapReductionOpportunity> {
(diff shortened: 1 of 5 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunities.java
```diff
@@ -99,21 +104,31 @@ public void visitDoStmt(DoStmt doStmt) {
...
-  private boolean isNonEmptyBlockStmtWithoutTopLevelDeclaration(Stmt stmt) {
-    if (!(stmt instanceof BlockStmt)) {
-      return false;
+    final BlockStmt childAsBlock = (BlockStmt) child;
+    if (childAsBlock.getNumStmts() == 0) {
+      // We do not unwrap an empty block statement, as there would be nothing to unwrap.
+      // Other reductions take responsibility for removing a redundant { }.
+      return;
+    }
+    // If unwrapping the child block may lead to name collisions then it cannot be done.
```
paulthomson: Could this cause an issue?

```c
int f()
{
  // child block:
  {
    int i;
  }
  // a block that comes later
  {
    int i;
  }
}
```

paulthomson: Never mind. Shadowing is fine.

paulthomson: I suppose this could cause issues, but this is not a new issue:

```c
void main() {
  {
      int max;
  }

  max(1.0, 1.0);
}

```

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunity.java
```diff
@@ -80,10 +86,36 @@ public void applyReductionImpl() {
...
   public boolean preconditionHolds() {
-    if (!(parentMap.getParent(wrapper) instanceof Stmt)) {
+    final IAstNode parentOfWrapper = parentMap.getParent(wrapper);
+    if (!(parentOfWrapper instanceof Stmt)) {
       return false;
     }
+    if (parentOfWrapper instanceof BlockStmt) {
+      // We need to make sure it is still the case that applying the unwrap will not lead to name
+      // clashes.
+      if (!Collections.disjoint(getNamesDeclaredDirectlyByBlock((BlockStmt) parentOfWrapper),
```
paulthomson: Does this logic match that of the finder? It looks slightly different. Maybe it does not need to be as precise because we only expect potential issues due to changes in the parent block?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunities.java
afd: Agreed; I have filed #408 for this.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/UnwrapReductionOpportunity.java
afd: I updated the comment to explain this a bit.

PR #403, opened 2019-03-11
Taught the constant folder how to turn e.g. int(1.0) into 1 and uint(…

afd:
…0.0) into 0u.

### DIFF
--- a/reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/FoldConstantReductionOpportunitiesTest.java
+++ b/reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/FoldConstantReductionOpportunitiesTest.java
@@ -549,6 +549,35 @@ public class FoldConstantReductionOpportunitiesTest {
         "void main() { 44200u; }");
   }
 
+  @Test
+  public void testCast1ToInt() throws Exception {
+    check("void main() { int(1.0); }", 1, "void main() { 1; }");
+  }
+
+  @Test
+  public void testCast0ToInt() throws Exception {
+    check("void main() { int(0.0); }", 1, "void main() { 0; }");
+  }
+
+  @Test
+  public void testCast100ToInt() throws Exception {
+    check("void main() { int(100.0); }", 1, "void main() { 100; }");
+  }
+
+  @Test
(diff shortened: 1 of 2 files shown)


reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/FoldConstantReductionOpportunitiesTest.java
```diff
@@ -549,6 +549,35 @@ public void testFoldHexAndOctalUInt() throws Exception {
...
+
+  @Test
+  public void testCast0ToUint() throws Exception {
+    check("void main() { uint(0.0); }", 1, "void main() { 0u; }");
+  }
+
+  @Test
+  public void testCast100ToUint() throws Exception {
+    check("void main() { uint(100.0); }", 1, "void main() { 100u; }");
+  }
```
paulthomson: E.g. `uint(X);` does not validate for me.

Could add some failure and tricky cases here. Could add several per test. I feel like some of these might fail badly, while others might fail but that's fine, although perhaps we should add them anyway to check that we don't get exceptions.

* `.0`
* `-.0`
* `1.1`
* `-1.1`
* `-0.0`
* `.1`
* `1.00001e1000000`
* `-1.0`
* `1.0e0`
* `-1.0e0`

afd: Good idea; I'll add this shader:

#version 310 es

void main() {
  uint(.0);
  int(-.0);
  int(1.1);
  int(-1.1);
  int(-0.0);
  int(.1);
  int(1.00001e1000000);
  uint(-1.0);
  uint(1.0e0);
  int(-1.0e0);
}

In a bunch of cases no constant folding will occur (e.g. due to - in negative cases being parsed as a unary operator, not as part of the literal), so I'll get the test to capture what we expect to work now, with a note that it could be updated if we made support more refined.

afd: OK, done.

PR #405, opened 2019-03-11
Added option RANDOM to glsl-to-spv-worker causing spirv-opt to be giv…

afd:
…en a random set of options on each invocation.

### DIFF
--- a/python/src/main/python/drivers/glsl-to-spv-worker.py
+++ b/python/src/main/python/drivers/glsl-to-spv-worker.py
@@ -16,6 +16,7 @@
 
 import argparse
 import os
+import random
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


python/src/main/python/drivers/glsl-to-spv-worker.py
```diff
@@ -402,7 +441,9 @@ def main():
 
     parser.add_argument(
         '--spirvopt',
-        help=runspv.SPIRV_OPT_OPTION_HELP)
+        help=runspv.SPIRV_OPT_OPTION_HELP + 'Pass RANDOM to have a random selection of '
```
paulthomson: Sounds a bit like you mean it will be random per shader job, but I don't think this is the case.

paulthomson: Should we maybe have a seed parameter so that we can e.g. restart a worker and keep the same random selection?

afd: The idea is that it should indeed be random per shader job (so no good for doing a reduction!).  I'm pretty sure that's what's implemented.

Good idea to add a seed parameter.  I'll do that, as well as documenting the issue with reductions being hard.

paulthomson: It’s a bit hard to see on GitHub, but it looks like the randomisation only happens near the start of main, not per incoming shader job.

afd: You were right!  Good catch.  Fixed now.

Issue #398, opened 2019-03-11
Represent 1 as the length of a normalized vector


afd:
In GraphicsFuzz we have various ways of generating 1, and a new one could be as length(normalize(some-non-zero-vector)).

afd:
Fixed via #413.

PR #412, opened 2019-03-14
Generate one from the length of normalized built-in function #398

jiradeto:
Implemented #398, add a new case in makeOpaqueZeroOrOne to make opaque zero or one from built-in GLSL functions and add a new method to generate 1 from the length of the normalized expression.


jiradeto:
close this issue since it's master branch base.

PR #423, opened 2019-03-20
Remove error listeners from parser to avoid console output

paulthomson:


paulthomson:
Yeah I had that, but read about how it is not very specific so decided to go with the other approach, even though it is not needed here; it is nice to be consistent.

Issue #427, opened 2019-03-21
Refactor function generating opaque zero and one


jiradeto:
As described in https://github.com/google/graphicsfuzz/pull/424#pullrequestreview-217342591, the makeOpqaueZeroOrOne function has more and more cases as we add new built-in functions, this method needs to be refactored. The code implemented to generate Zero and One should be separated.

afd:
Fixed via #447.

PR #447, opened 2019-04-01
Seperate functions making opaque 1 and 0

jiradeto:
Related issue #427, built-in functions used to generate opaque zero and one expressions are now separated into the different methods.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -127,12 +127,73 @@ public final class OpaqueExpressionGenerator {
 
   public Expr makeOpaqueZero(BasicType type, boolean constContext, final int depth,
         Fuzzer fuzzer) {
-    return makeOpaqueZeroOrOne(true, type, constContext, depth, fuzzer);
+    if (isTooDeep(depth)) {
+      return makeLiteralZeroOrOne(true, type);
+    }
+    final int newDepth = depth + 1;
+    while (true) {
+      final int numTypesOfZero = 3;
+      switch (generator.nextInt(numTypesOfZero)) {
+        case 0:
+          // represent 0 as the natural logarithm of opaqueOne, e.g. log(1.0)
+          if (!BasicType.allGenTypes().contains(type)) {
+            continue; // log doesn't operate on non-gen types.
+          }
+          return new FunctionCallExpr("log", makeOpaqueOne(type, constContext, newDepth,
+              fuzzer));
+        case 1:
+          // represent 0 as sin(opaqueZero) function, e.g. sin(0.0)
+          if (!BasicType.allGenTypes().contains(type)) {
+            continue; // sin doesn't operate on non-gen types.
+          }
+          return new FunctionCallExpr("sin", makeOpaqueZero(type, constContext, newDepth,
+              fuzzer));
+        case 2:
+          return makeOpaqueZeroOrOne(true, type, constContext, newDepth, fuzzer);
+        default:
+          throw new RuntimeException();
+      }
+    }
   }
 
   public Expr makeOpaqueOne(BasicType type, boolean constContext, final int depth,
         Fuzzer fuzzer) {
-    return makeOpaqueZeroOrOne(false, type, constContext, depth, fuzzer);
+
+    if (isTooDeep(depth)) {
+      return makeLiteralZeroOrOne(false, type);
+    }
+    final int newDepth = depth + 1;
+    while (true) {
+      final int numTypesOfOne = 4;
+      switch (generator.nextInt(numTypesOfOne)) {
+        case 0:
+          // represent 1 as the length of normalized non-zero vector
+          if (!BasicType.allGenTypes().contains(type)) {
+            continue; // normalize doesn't operate on non-gen types.
+          }
+          Expr normalizedExpr = new FunctionCallExpr("normalize", makeOpaqueOne(
+              type, constContext, newDepth, fuzzer));
+          return new FunctionCallExpr("length", normalizedExpr);
+        case 1:
+          // represent 1 as cos(opaqueZero) function, e.g. cos(0.0)
+          if (!BasicType.allGenTypes().contains(type)) {
+            continue; // cos doesn't operate on non-gen types.
+          }
+          return new FunctionCallExpr("cos", makeOpaqueZero(type, constContext, newDepth,
+              fuzzer));
+        case 2:
+          // represent 1 as the exponential function of opaqueZero, e.g. exp(0.0)
+          if (!BasicType.allGenTypes().contains(type)) {
+            continue; // exp doesn't operate on non-gen types.
+          }
+          return new FunctionCallExpr("exp", makeOpaqueZero(type, constContext, newDepth,
+              fuzzer));
+        case 3:
+          return makeOpaqueZeroOrOne(false, type, constContext, newDepth, fuzzer);
+        default:
+          throw new RuntimeException();
+      }
+    }
   }
 
   private Expr makeOpaqueZeroOrOne(boolean isZero, BasicType type, boolean constContext,
@@ -145,16 +206,16 @@ public final class OpaqueExpressionGenerator {
     }
     final int newDepth = depth + 1;
     while (true) {
-      final int numTypesOfZeroOrOne = 8;
+      final int numTypesOfZeroOrOne = 3;
       switch (generator.nextInt(numTypesOfZeroOrOne)) {
         case 0:
           // Make an opaque value recursively and apply an identity function to it
           return applyIdentityFunction(makeOpaqueZeroOrOne(isZero, type, constContext,
-                newDepth, fuzzer),
-                type,
-                constContext,
-                newDepth,
-                fuzzer);
+              newDepth, fuzzer),
+              type,
+              constContext,
+              newDepth,
+              fuzzer);
         case 1:
           // injectionSwitch.x or injectionSwitch.y
           if (constContext || !generationParams.getInjectionSwitchIsAvailable()) {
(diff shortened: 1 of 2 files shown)


jiradeto:
@afd 

If I understand you correctly, "Factor out each way into a method" means having a class for each way to make zero or one that implements the functional interface `OpaqueZeroOneFactory` and in each class, the override method tryMakeOpaque would return `Optional.empty()` when the case we currently continue, otherwise return the `Optional.of( FunctionCallExpr )`?
  
I uploaded the example implementation based on your comment in this [link](https://gist.github.com/jiradeto/9de87afedab484e1307b6672315e3ff0). Please kindly help to review. 

Thank you very much

paulthomson:
Just to let you know: afd is away until next week. I’m sure he will give some advice and feedback then.

jiradeto:
@paulthomson thank you for informing me, all my doubts are clear now after digging through the code.

jiradeto:
@afd Thank you so much! your detailed explanation is really helpful.
Code has been fixed. 

> The problem I see with this is that it skews probabilities. Take the case of making zero. We'll have a bunch of ways to make just zero - say M of them, and a bunch of ways to make zero that can also make one - say N of them. With your design, we we end up choosing each of the ways to only make zero with probability 1/(M+1). With probability 1/(M+1) we decide to choose one of the options to make zero that could also make 1, and there are N of those, so each has overall probability 1/N(M+1) of being chosen.
> 
> I think we should take a different approach, where basically we have a list of method references of functions that can create zero, and a list of method references of functions that can create one (with overlap between the lists for functions that can create both), and then sample uniformly from a given list.
> 
> More concretely, something like the following:
> 
> * Define a functional interface, OpaqueZeroOneFactory, with a single method:
>   Optional tryMakeOpaque(BasicType type, boolean constContext, final int depth,
>   Fuzzer fuzzer, boolean isZero); // Optional.empty() gets returned in some of the cases where we would currently continue
> * Have (possibly static) methods:
>   
>   * waysToMakeZero - returns a list of OpaqueZeroOneFactory, with all the factories that can make zero
>   * waysToMakeOne - returns a list of OpaqueZeroOneFactory, with all the factories that can make one
> * Factor out each way to make zero into a method that has the right signature to be an OpaqueZeroOneFactory and that asserts that its isZero parameter is true.  Similar for one.  For the mechanisms that can make both zero and one, the isZero parameter is used accordingly.  Each of these methods will call makeOpaqueZeroOrOne (see below) to keep the recursion that we currently have.
> * Have a single method, makeOpaqueZeroOrOne, that (a) checks whether recursion is too deep and returns something simple in that case (as is currently done), and (b) otherwise tests isZero, gets the list of possible factory functions, and repeatedly samples from this until it finds one that returns an Optional that is present, and ultimately returns the Expr in that optional.
> 
> Let me know what you think.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
 
   }
 
+  @FunctionalInterface
+  public interface OpaqueZeroOneFactory {
```
afd: Can you please move this interface into its own file?

And can you move the static methods for creating the lists of factories into OpaqueExpressionGenerator, and make them be non-static, rather than having them be static methods of the interface?

See comment below for (one reason) why this is a good idea.  In general I try to avoid an interface knowing about its implementing classes.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    Optional tryMakeOpaque(BasicType type, boolean constContext, final int depth,
+                           Fuzzer fuzzer, boolean isZero);
+
```
afd: I suggest having a method "waysToMakeZeroOrOne" that returns the list of common ways, and having each of "waysToMakeZero" and "waysToMakeOne" return a list comprised of the common ways, plus any additional more specific ways.  This will avoid the problem of forgetting to add a new common way in both places.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
 
   }
 
+  @FunctionalInterface
+  public interface OpaqueZeroOneFactory {
+    Optional tryMakeOpaque(BasicType type, boolean constContext, final int depth,
```
afd: Change Optional to Optional<Expr> (in general, avoid raw generic types)

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+          opaqueExpressionGenerator::opaqueZeroOrOneFromInjectionSwitch,
+          opaqueExpressionGenerator::opaqueZeroOrOneSquareRoot,
+          opaqueExpressionGenerator::opaqueOneExponential,
+          opaqueExpressionGenerator::opaqueOneCosine,
+          opaqueExpressionGenerator::opaqueOneNormalizedVectorLength
+      );
+    }
+  }
+
+  private Optional opaqueZeroOrOneFromIdentityFunction(BasicType type, boolean constContext,
```
afd: Update the return type of this, and all similar methods, to Optional<Expr>.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      return Optional.empty();
+    }
+    return Optional.of(new FunctionCallExpr("sqrt", makeOpaqueZeroOrOne(isZero, type, constContext,
+        depth, fuzzer)));
+  }
+
+  private Optional opaqueZeroSin(BasicType type, boolean constContext, final int depth,
+                                 Fuzzer fuzzer, boolean isZero) {
+    // represent 0 as sin(opaqueZero) function, e.g. sin(0.0)
+    if (!isZero) {
```
afd: I would prefer to assert !isZero here.  I don't think the function should ever be called if isZero is false, so an assertion would be more appropriate.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      return Optional.empty();
+    }
+    if (!BasicType.allGenTypes().contains(type)) {
+      return Optional.empty();
```
afd: In contrast to my remark about the assertion above, it is indeed appropriate to return Optional.empty() here, since we may well call this method with an incompatible type.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -84,6 +85,134 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    }
+    return Optional.of(new FunctionCallExpr("sin", makeOpaqueZero(type, constContext, depth,
+        fuzzer)));
+  }
+
+  private Optional opaqueZeroLogarithm(BasicType type, boolean constContext, final int depth,
+                                       Fuzzer fuzzer, boolean isZero) {
+    // represent 0 as the natural logarithm of opaqueOne, e.g. log(1.0)
+    if (!isZero) {
```
afd: Similar comment about assertion - please apply to all relevant places.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -120,109 +249,44 @@ private Expr applyBinaryIdentityFunction(Expr expr, Expr opaqueExpr, BinOp opera
...
-              fuzzer));
-        default:
-          throw new RuntimeException();
-      }
-    }
+
+    // If isZero holds, we are making an opaque zero, otherwise an opaque one
+    final List<OpaqueZeroOneFactory> opaqueFactories = isZero
+        ? OpaqueZeroOneFactory.waysToMakeZero(this)
+        : OpaqueZeroOneFactory.waysToMakeOne(this);
```
afd: Passing in "this" to "waysToMakeZero" and "waysToMakeOne" is not very nice.  By moving these methods out of the interface and into OpaqueExpressionGenerator, and by making them instance methods rather than static methods, you avoid this problem.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -120,109 +249,44 @@ private Expr applyBinaryIdentityFunction(Expr expr, Expr opaqueExpr, BinOp opera
...
+    final List<Optional> availableOpaqueFactories = opaqueFactories
+        .stream()
+        .map(item -> item.tryMakeOpaque(type, constContext, newDepth, fuzzer, isZero))
+        .filter(Optional::isPresent)
+        .collect(Collectors.toList());
+
+    final Expr expr = (Expr) availableOpaqueFactories
```
afd: The cast to Expr will not be necessary once you change the factory so that it returns Optional<Expr>.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -120,109 +249,44 @@ private Expr applyBinaryIdentityFunction(Expr expr, Expr opaqueExpr, BinOp opera
...
+
+    // If isZero holds, we are making an opaque zero, otherwise an opaque one
+    final List<OpaqueZeroOneFactory> opaqueFactories = isZero
+        ? OpaqueZeroOneFactory.waysToMakeZero(this)
+        : OpaqueZeroOneFactory.waysToMakeOne(this);
+    final List<Optional> availableOpaqueFactories = opaqueFactories
+        .stream()
+        .map(item -> item.tryMakeOpaque(type, constContext, newDepth, fuzzer, isZero))
+        .filter(Optional::isPresent)
+        .collect(Collectors.toList());
```
afd: The above is quite clever, but I do not like it ;-)  In particular, I don't like the strategy of trying to make an opaque expression to see whether it will be possible.

I would prefer something along the following lines, which loops forever until it finds something that works:

while (true) {
  final OpaqueZeroOneFactory factory = opaqueFactories.get(random index into opaqueFactories);
  final Optional<Expr> maybeResult = factory.tryMakeOpaque(...);
  if (maybeResult.isPresent()) {
    return maybeResult.get();
  }
}

By construction, there will always be at least one case that will work (injectionSwitch), so the loop will probabilistically terminate.

jiradeto: Thank you! your code looks really clean and more understandable. Just out of my curiosity, why do we need `while (true)` in this case?

afd: The while (true) means that we keep on trying until we find a factory from which we get a non-empty optional (which we will eventually).

An alternative would be to loop while the list of factories is non-empty, and remove each unsuccessful factory from the list right after finding out the optional is not present. If the loop would terminate - indicating that nothing worked - a RuntimeException with appropriate error string should be thrown (rather than returning null) since we should always find a suitable factory I believe.

jiradeto: The alternative approach is reasonable for me. The code has been fixed!
Thank you again for your informative comment.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueZeroOneFactory.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2018 The GraphicsFuzz Project Authors
```
afd: Can you change this to 2019?

jiradeto: Done!

Issue #397, opened 2019-03-11
Represent 0 as the length of a zero vector


afd:
In GraphicsFuzz we have various ways of generating 0, and a new one could be as len(some-vector-guaranteed-to-be-zero).

afd:
Fixed via #472

PR #448, opened 2019-04-04
Add possibility to represent zero as length of zero vector

niazarak:
Related issue (#397), (#412 )
Implemented in the same way we make opaque "one" with length of normalized vectors.


niazarak:
Uh oh, didn't see that #432 PR, closing this one

Issue #450, opened 2019-04-07
Allow random generation of reference input data in generated shader job


abelbriggs1:
Currently, when generating a family of variants for a shader, the generator simply copies the original shader job JSON into the new directory and uses it to create the variant JSON files, like so:

bubblesort_flag.json
`{
    "injectionSwitch": {
        "func":  "glUniform2f",
        "args": [
            0.0,
            1.0
        ]
    },
    "resolution": {
        "func":  "glUniform2f",
        "args": [
            256.0,
            256.0
        ]
    }
}`

reference.json
`{
  "injectionSwitch": {
    "func": "glUniform2f",
    "args": [
      0.0,
      1.0
    ]
  },
  "resolution": {
    "func": "glUniform2f",
    "args": [
      256.0,
      256.0
    ]
  }
}`

This means that no matter how many variants you generate, the input uniforms will never change unless you use a different original JSON file. This is great in most cases, but some shaders/algorithms could have special cases depending on the input values that could result in different behavior. Covering the different behaviors of a shader and its variants with different inputs would increase the possibility of finding compiler bugs.

It would be nice to be able to tell the generator to randomize some inputs in our JSON file while preparing it as a reference job. For example, a string that would tell the type of random value and its upper bound.

Perhaps something like this: (notice the first data field)

comp-0003-random-middle-square.json
`{
  "$compute": {
    "num_groups": [ 2, 1, 1 ],
    "buffer": {
      "binding": 0,
      "fields":
      [
        { "type": "int", "data": [ "randint(12345)" ] },
        { "type": "int", "data":
          [ 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0
          ]
        }
      ]
    }
  }
}`

reference.json
`{
  "$compute": {
    "num_groups": [ 2, 1, 1 ],
    "buffer": {
      "binding": 0,
      "fields":
      [
        { "type": "int", "data": [ 7890 ] },
        { "type": "int", "data":
          [ 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0
          ]
        }
      ]
    }
  }
}`

[...]

PR #455, opened 2019-04-13
Allow random generation of reference input data in generated shader job

abelbriggs1:
Fixes/implements #450.


afd:
Thanks for making a start on this!  As mentioned in my comment on issue #450, I think we should approach it a different way, and I think it would thus be best to close this PR in due course and open a different one for the alternative approach, though I suspect you can re-used some of the code in this PR.

Issue #110, opened 2018-11-17
Reducer: reduce double brackets (( )) to simple ones ( ) after final _GLF_ macro replacement


hevrard:
This:

`if ( _GLF_BLAH((X)) ) {}`

may lead to this at the very end of reduction, when _GLF_ macros are replaced by their expension:

`if ((X)) {}`

But this double brackets could be reduced to:

`if (X) {}`

afd:
In #471 I have added a test to illustrate this problem.  When the issue is fixed, that test should pass.  (It would be worth adding a few others like it.)

jiradeto:
Fixed by #489.

PR #489, opened 2019-05-17
Remove parentheses after eliminating macros

jiradeto:
Related issue: #110 

- Remove the unneeded parentheses that exist after the macros elimination
- Update statement in the test cases that previously do not have the wrapper

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
@@ -33,23 +33,23 @@ public class EliminateInjectionMacrosVisitor extends StandardVisitor {
   protected void visitChildFromParent(IAstNode child, IAstNode parent) {
     super.visitChildFromParent(child, parent);
     if (child instanceof FunctionCallExpr) {
-      Expr expr = (Expr) child;
-      if (MacroNames.isIdentity(expr)
-          || MacroNames.isZero(expr)
-          || MacroNames.isOne(expr)
-          || MacroNames.isFalse(expr)
-          || MacroNames.isTrue(expr)) {
+      final FunctionCallExpr functionCallExpr = (FunctionCallExpr) child;
+      if (MacroNames.isIdentity(functionCallExpr)
+          || MacroNames.isZero(functionCallExpr)
+          || MacroNames.isOne(functionCallExpr)
+          || MacroNames.isFalse(functionCallExpr)
+          || MacroNames.isTrue(functionCallExpr)) {
         parent.replaceChild(child,
-            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(1)));
-      } else if (MacroNames.isFuzzed(expr)
-          || MacroNames.isDeadByConstruction(expr)
-          || MacroNames.isSwitch(expr)
-          || MacroNames.isLoopWrapper(expr)
-          || MacroNames.isIfWrapperFalse(expr)
-          || MacroNames.isIfWrapperTrue(expr)
+            addParenthesesIfNecessary(parent, functionCallExpr.getChild(1)));
+      } else if (MacroNames.isFuzzed(functionCallExpr)
+          || MacroNames.isDeadByConstruction(functionCallExpr)
+          || MacroNames.isSwitch(functionCallExpr)
+          || MacroNames.isLoopWrapper(functionCallExpr)
+          || MacroNames.isIfWrapperFalse(functionCallExpr)
+          || MacroNames.isIfWrapperTrue(functionCallExpr)
       ) {
         parent.replaceChild(child,
-            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(0)));
+            addParenthesesIfNecessary(parent, functionCallExpr.getChild(0)));
       }
     }
   }
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
@@ -16,9 +16,12 @@
 
 package com.graphicsfuzz.reducer.glslreducers;
 
+import com.graphicsfuzz.common.ast.IAstNode;
 import com.graphicsfuzz.common.ast.decl.ScalarInitializer;
 import com.graphicsfuzz.common.ast.expr.ArrayIndexExpr;
+import com.graphicsfuzz.common.ast.expr.BinOp;
 import com.graphicsfuzz.common.ast.expr.BinaryExpr;
+import com.graphicsfuzz.common.ast.expr.ConstantExpr;
 import com.graphicsfuzz.common.ast.expr.Expr;
 import com.graphicsfuzz.common.ast.expr.FunctionCallExpr;
 import com.graphicsfuzz.common.ast.expr.MemberLookupExpr;
@@ -26,10 +29,14 @@ import com.graphicsfuzz.common.ast.expr.ParenExpr;
 import com.graphicsfuzz.common.ast.expr.TernaryExpr;
 import com.graphicsfuzz.common.ast.expr.TypeConstructorExpr;
 import com.graphicsfuzz.common.ast.expr.UnaryExpr;
+import com.graphicsfuzz.common.ast.expr.VariableIdentifierExpr;
 import com.graphicsfuzz.common.ast.stmt.DoStmt;
+import com.graphicsfuzz.common.ast.stmt.ExprStmt;
 import com.graphicsfuzz.common.ast.stmt.ForStmt;
 import com.graphicsfuzz.common.ast.stmt.IfStmt;
 import com.graphicsfuzz.common.ast.stmt.LoopStmt;
+import com.graphicsfuzz.common.ast.stmt.SwitchStmt;
+import com.graphicsfuzz.common.ast.stmt.WhileStmt;
 import com.graphicsfuzz.common.ast.visitors.StandardVisitor;
 import com.graphicsfuzz.reducer.reductionopportunities.MacroNames;
 
@@ -90,6 +97,35 @@ public class EliminateInjectionMacrosVisitor extends StandardVisitor {
     }
   }
 
+  @Override
+  public void visitSwitchStmt(SwitchStmt switchStmt) {
+    super.visitSwitchStmt(switchStmt);
+    if (MacroNames.isSwitch(switchStmt.getExpr())) {
+      switchStmt.setExpr(switchStmt.getExpr().getChild(0));
+    }
+  }
+
+  @Override
+  public void visitWhileStmt(WhileStmt whileStmt) {
+    super.visitWhileStmt(whileStmt);
+    cleanupLoop(whileStmt);
+  }
+
+  @Override
+  public void visitExprStmt(ExprStmt exprStmt) {
+    super.visitExprStmt(exprStmt);
+    if (MacroNames.isIdentity(exprStmt.getExpr())
+        || MacroNames.isZero(exprStmt.getExpr())
+        || MacroNames.isOne(exprStmt.getExpr())
+        || MacroNames.isFalse(exprStmt.getExpr())
+        || MacroNames.isTrue(exprStmt.getExpr())) {
+      exprStmt.setExpr(exprStmt.getExpr().getChild(1));
+    } else if (MacroNames.isFuzzed(exprStmt.getExpr())
+        || MacroNames.isDeadByConstruction(exprStmt.getExpr())) {
+      exprStmt.setExpr(exprStmt.getExpr().getChild(0));
+    }
+  }
+
   private void cleanupLoop(LoopStmt loopStmt) {
     if (MacroNames.isLoopWrapper(loopStmt.getCondition())) {
       loopStmt.setCondition(loopStmt.getCondition().getChild(0));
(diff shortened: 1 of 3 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -129,15 +165,58 @@ private void cleanUpMacros(Expr parent) {
...
       scalarInitializer.setExpr(scalarInitializer.getExpr().getChild(0));
     }
   }
 
   private void replaceChildWithGrandchild(Expr parent, int childIndex, int grandchildIndex) {
     assert parent.getChild(childIndex).getNumChildren() > grandchildIndex;
-    parent
-        .setChild(childIndex, new ParenExpr(parent.getChild(childIndex).getChild(grandchildIndex)));
+
+    if (suitableToRemoveParentheses(parent, parent.getChild(childIndex))) {
```
afd: Rather than deciding whether to remove parentheses, we are in fact deciding whether to add them.  So please can you rename the "suitableToRemoveParentheses" method along those lines?  It should check whether it is necessary to add them.

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -129,15 +165,58 @@ private void cleanUpMacros(Expr parent) {
...
+          new ParenExpr(parent.getChild(childIndex).getChild(grandchildIndex)));
+    }
+  }
+
+  private boolean suitableToRemoveParentheses(IAstNode parent, Expr child) {
+
+    if (child instanceof ConstantExpr
+        || child instanceof VariableIdentifierExpr
+        || child instanceof FunctionCallExpr) {
+      // It's fine to remove parentheses in cases such as (5), (x) and (sin(a)).
```
afd: When you change the method name (as per the previous comment), be sure to update the comments accordingly.

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -129,15 +165,58 @@ private void cleanUpMacros(Expr parent) {
...
+
+    if (!(parent instanceof Expr)) {
+      // These are outer-most parentheses; fine to remove them.
+      return true;
+    }
+    if (parent instanceof ParenExpr) {
+      // These are parentheses within parentheses; fine to remove them.
+      return true;
+    }
+    if (parent instanceof FunctionCallExpr) {
```
afd: As the comma operator cannot appear directly as a macro argument, you should instead assert that the child is *not* an instance of the comma operator.

That would be clearer, because having code to deal with the comma operator suggests that we might expect the operator here, when in fact we do not.

reducer/src/test/java/com/graphicsfuzz/reducer/util/SimplifyTest.java
```diff
@@ -55,30 +51,28 @@ public void testWhileParenthesesRemoved() throws Exception {
     CompareAsts.assertEqualAsts(expected, simplifiedTu);
   }
 
-  @Ignore
   @Test
   public void testForParenthesesRemoved() throws Exception {
     final TranslationUnit tu = ParseHelper.parse("void main() {"
         + " for ("
-        + "   int i = _GLF_ONE(1, _GLF_IDENTITY(1, 1));"
+        + "   int i = int(_GLF_ONE(1.0, _GLF_IDENTITY(1.0, 1.0)));"
```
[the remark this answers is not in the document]
jiradeto: Just for consistency, I noticed that most for-loop variables in variant shaders generated by GraphicsFuzz are float number that wrapped with `int(xx)`.  Does it make sense to do so?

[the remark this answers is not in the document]
afd: It doesn't matter either way.

Fine to change to the int(...) case if you wish.  As a rule of thumb, it's better to do less in a single PR though.  If you want to do a refactoring, or improve consistency, or whatever, it's fine to do that in a separate PR.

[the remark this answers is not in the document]
jiradeto: Done, thank you!

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/test/java/com/graphicsfuzz/reducer/util/SimplifyTest.java
```diff
@@ -40,11 +37,10 @@ public void testIfParenthesesRemoved() throws Exception {
     CompareAsts.assertEqualAsts(expected, simplifiedTu);
   }
 
-  @Ignore
   @Test
   public void testWhileParenthesesRemoved() throws Exception {
     final TranslationUnit tu = ParseHelper.parse("void main() {"
-        + "  while (_GLF_DEAD(_GLF_FALSE(false, false))) {"
+        + "  while(_GLF_WRAPPED_LOOP(_GLF_FALSE(false ,_GLF_DEAD(_GLF_FALSE(false, false))))) {"
```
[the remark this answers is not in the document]
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

jiradeto:
Thanks! redundant methods were removed.

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,84 @@
...
+            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(1)));
+      } else if (MacroNames.isFuzzed(expr)
+          || MacroNames.isDeadByConstruction(expr)
+          || MacroNames.isSwitch(expr)
+          || MacroNames.isLoopWrapper(expr)
+          || MacroNames.isIfWrapperFalse(expr)
+          || MacroNames.isIfWrapperTrue(expr)
+      ) {
+        parent.replaceChild(child,
+            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(0)));
```
afd: Similar

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,84 @@
...
-  public void visitIfStmt(IfStmt ifStmt) {
-    super.visitIfStmt(ifStmt);
-    if (MacroNames.isDeadByConstruction(ifStmt.getCondition())
-        || MacroNames.isIfWrapperFalse(ifStmt.getCondition())
-        || MacroNames.isIfWrapperTrue(ifStmt.getCondition())) {
-      ifStmt.setCondition(ifStmt.getCondition().getChild(0));
+  protected void visitChildFromParent(IAstNode child, IAstNode parent) {
+    super.visitChildFromParent(child, parent);
+    if (child instanceof FunctionCallExpr) {
+      Expr expr = (Expr) child;
```
afd: Make expr final (in general try to always use final unless there's a reason why you cannot).

Change Expr to FunctionCallExpr.  The functions like MacroNames.isZero(...) will accept arguments of type FunctionCallExpr, as it is a subclass of Expr, and then you can use expr instead of (FunctionCallExpr) child below.

I also suggest that you call the variable functionCallExpr rather than expr is you change its type to FunctionCallExpr.

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,84 @@
...
+      if (MacroNames.isIdentity(expr)
+          || MacroNames.isZero(expr)
+          || MacroNames.isOne(expr)
+          || MacroNames.isFalse(expr)
+          || MacroNames.isTrue(expr)) {
+        parent.replaceChild(child,
+            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(1)));
```
afd: Change (FunctionCallExpr) child to functionCallExpr (see above comment)

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,84 @@
...
+        parent.replaceChild(child,
+            addParenthesesIfNecessary(parent, ((FunctionCallExpr) child).getChild(0)));
+      }
     }
   }
 
-  private void cleanupLoop(LoopStmt loopStmt) {
-    if (MacroNames.isLoopWrapper(loopStmt.getCondition())) {
-      loopStmt.setCondition(loopStmt.getCondition().getChild(0));
+  private IAstNode addParenthesesIfNecessary(IAstNode parent, Expr child) {
```
afd: This method is so short that I don't think it makes sense to have a separate method, "ifParenthesesNecessary".  Can you just inline the logic for "ifParenthesesNecessary" into this method?

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,84 @@
...
-  public void visitScalarInitializer(ScalarInitializer scalarInitializer) {
-    super.visitScalarInitializer(scalarInitializer);
-    if (MacroNames.isFuzzed(scalarInitializer.getExpr())) {
-      scalarInitializer.setExpr(scalarInitializer.getExpr().getChild(0));
+    if (parent instanceof FunctionCallExpr) {
+      if (child instanceof BinaryExpr) {
+        // The binary operator must not be the comma operator
+        // as it cannot appear directly as a macro argument.
+        assert ((BinaryExpr) child).getOp() != BinOp.COMMA;
+        return true;
```
afd: I do not understand why you return true here.  Let's say the parent is a function call expression, e.g. foo, and the child is a binary expression, e.g. a + b.

We thus have something like:

foo(..., a + b, ...)

there is no need to put parentheses around a + b.

I suggest replacing this whole conditional with:

assert (!(child instanceof BinaryExpr) || ((BinaryExpr) child).getOp() != BinOp.COMMA);

and keeping the nice comment about the reason for the comma operator.

But I believe false can be returned whenever parent is a FunctionCallExpr.

afd: Can you add an additional test that for the above case, e.g. something that checks that:

foo(_GLF_IDENTITY(1 + 2, 1 + 2))

should get simplified to:

foo(1 + 2)
?

You should find that the test fails with your current code, and after you address the above comment it should pass.

jiradeto: Got it! thank you for a very clear explanation.

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/glslreducers/EliminateInjectionMacrosVisitor.java
```diff
@@ -16,128 +16,75 @@
...
+      return child;
     }
-  }
 
-  @Override
-  public void visitScalarInitializer(ScalarInitializer scalarInitializer) {
-    super.visitScalarInitializer(scalarInitializer);
-    if (MacroNames.isFuzzed(scalarInitializer.getExpr())) {
-      scalarInitializer.setExpr(scalarInitializer.getExpr().getChild(0));
+    if (parent instanceof FunctionCallExpr) {
```
afd: For completeness, could you add something like:

      // If parent is a function call, adding a new parentheses would be redundant,
      // e.g. foo(..., _GLF_FUNCTION(a + b), ...).

?

I really like the other comments you have added like this, so it would be great to do the same in this case too.

jiradeto: Done!

Issue #478, opened 2019-05-16
reducer: try removing uniforms from JSON if they are not defined in the shader


paulthomson:

afd:
#481 provides tests to guide in addressing this issue.

Intuitively, the problem is that during test case reduction, the reducer might remove uniform declarations from a shader but will fail to remove associated metadata about those uniforms from the PipelineInfo object that accompanies a set of shaders.  It would be good to teach the reducer how to do this.

This will involve creating two new classes:

- RemoveRedundantUniformMetadatReductionOpportunity
- RemoveRedundantUniformMetadatReductionOpportunities
(see similar classes for inspiration as to what to do in them).

It will also require:

- adding a new "finder" method in IReductionOpportunityFinder.
- adding an invocation of this method in ReductionOpportunities::getReductionOpportunities
- adding an invocation of this method to populate cleanupPasses in ReductionDriver

afd:
Fixed via #490.

PR #490, opened 2019-05-20
Reducer opportunity to remove redundant uniform metadata

jiradeto:
Related issue: #478

Added a new reducer opportunity which removes all redundant uniforms metadata in JSON except for the uniforms that are declared but not used.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
@@ -1011,7 +1011,14 @@ public class ShaderJobFileOperations {
     byte[] computeData = isFile(computeShaderFile)
         ? readFileToByteArray(computeShaderFile)
         : new byte[0];
-    byte[] combinedData = new byte[vertexData.length + fragmentData.length + computeData.length];
+    // This metadata is required since in the case of
+    // RemoveRedundantUniformMetadataReductionOpportunities
+    // only metadata in json is removed, not the shaders.
+    byte[] metaData = isFile(shaderJobFile)
+        ? readFileToByteArray(shaderJobFile)
+        : new byte[0];
+    byte[] combinedData =
+        new byte[vertexData.length + fragmentData.length + computeData.length + metaData.length ];
     System.arraycopy(
         vertexData,
         0,
@@ -1030,6 +1037,12 @@ public class ShaderJobFileOperations {
         combinedData,
         vertexData.length + fragmentData.length,
         computeData.length);
+    System.arraycopy(
+        metaData,
+        0,
+        combinedData,
+        vertexData.length + fragmentData.length + computeData.length,
+        metaData.length);
     return DigestUtils.md5Hex(combinedData);
   }
 
--- a/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
@@ -1011,9 +1011,8 @@ public class ShaderJobFileOperations {
     byte[] computeData = isFile(computeShaderFile)
         ? readFileToByteArray(computeShaderFile)
         : new byte[0];
-    // This metadata is required since in the case of
-    // RemoveRedundantUniformMetadataReductionOpportunities
-    // only metadata in json is removed, not the shaders.
+    //  This metadata is required in order to distinguish between shader jobs
+    //  with identical shaders but different pipeline information.
     byte[] metaData = isFile(shaderJobFile)
         ? readFileToByteArray(shaderJobFile)
         : new byte[0];
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
@@ -19,76 +19,48 @@ package com.graphicsfuzz.reducer.reductionopportunities;
 import com.graphicsfuzz.common.ast.TranslationUnit;
 import com.graphicsfuzz.common.ast.decl.VariableDeclInfo;
 import com.graphicsfuzz.common.ast.decl.VariablesDeclaration;
-import com.graphicsfuzz.common.ast.expr.VariableIdentifierExpr;
+import com.graphicsfuzz.common.ast.type.TypeQualifier;
+import com.graphicsfuzz.common.ast.visitors.StandardVisitor;
+import com.graphicsfuzz.common.ast.visitors.VisitationDepth;
 import com.graphicsfuzz.common.transformreduce.ShaderJob;
-import com.graphicsfuzz.common.util.ListConcat;
-import com.graphicsfuzz.common.util.PipelineInfo;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
-public class RemoveRedundantUniformMetadataReductionOpportunities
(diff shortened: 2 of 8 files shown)


jiradeto:
@afd, the test case below `testReductionWithUniformBindings` in ReductionDriverTest failed due to the assertion did not find any uniform binding since they were removed by the new opportunity.

https://github.com/google/graphicsfuzz/blob/aa0ce2ff6dab59329325d5df90ab7ae902112010/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java#L299


Does the `RemoveRedundantUniformMetadataReductionOpportunities` need to check whether the uniform binding is required before start finding any opportunity? i.e., if a uniform binding is required, the finder will return empty opportunities and the uniform metadata will not be removed

afd:
@jiradeto This is an interesting edge case that I had not foreseen!

The tool doesn't properly support uniform bindings, which are required for Vulkan testing.  To work around lack of support, they get added temporarily each time a shader is written to a file.  But if there are no uniforms then they don't need to be added.

I suggest that you change the assertion to:

      assert state.getPipelineInfo().getNumUniforms() == 0 || state.hasUniformBindings();

This should solve the problem for now.

(In time it would be good for the tool to properly support uniform bindings, so that this business of adding them and removing them could be dispensed with, but that's not for this issue.)

jiradeto:
@afd, thank you! the assertion is updated and also the shader job metadata is included as the part of MD5 hash to prevent the reduction loop.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
```diff
@@ -0,0 +1,94 @@
...
+import com.graphicsfuzz.common.ast.expr.VariableIdentifierExpr;
+import com.graphicsfuzz.common.transformreduce.ShaderJob;
+import com.graphicsfuzz.common.util.ListConcat;
+import com.graphicsfuzz.common.util.PipelineInfo;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class RemoveRedundantUniformMetadataReductionOpportunities
+    extends ReductionOpportunitiesBase<RemoveRedundantUniformMetadataReductionOpportunity> {
```
afd: I am unsure whether this class should extend ReductionOpportunitiesBase after all, since it is fundamentally a bit different: it may need to consider multiple shaders in order to know what the opportunities are (since a uniform might be used in a vertex shader but not fragment shader, for example).

I think it might be preferable to have a class that implements IReductionOpportunity directly.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunitiesTest.java
```diff
@@ -82,14 +74,12 @@ public void testDoNotRemoveUsed() throws Exception {
     assertEquals(1, shaderJob.getPipelineInfo().getNumUniforms());
 
     // There should be exactly one opportunity to remove a piece of unused pipeline state.
```
afd: There was a typo in this comment; it should read "There should be no opportunities to ..." - please fix in this PR.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
```diff
@@ -0,0 +1,94 @@
...
+    // This prevents removing a uniform declared but not used.
+    super.visitVariablesDeclaration(variablesDeclaration);
+    for (VariableDeclInfo vdi : variablesDeclaration.getDeclInfos()) {
+      removeRedundantUniformName(vdi.getName());
+    }
+  }
+
+  @Override
+  public void visitVariableIdentifierExpr(VariableIdentifierExpr variableIdentifierExpr) {
+    super.visitVariableIdentifierExpr(variableIdentifierExpr);
```
afd: This should not be needed.  We cannot use a variable if we have not declared it, and if we tag uniforms as being needed if they are declared then we are fine.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
```diff
@@ -0,0 +1,94 @@
...
+    finder.visit(tu);
+    return finder.getOpportunities();
+  }
+
+  @Override
+  public void visitVariablesDeclaration(VariablesDeclaration variablesDeclaration) {
+    // This prevents removing a uniform declared but not used.
+    super.visitVariablesDeclaration(variablesDeclaration);
+    for (VariableDeclInfo vdi : variablesDeclaration.getDeclInfos()) {
+      removeRedundantUniformName(vdi.getName());
```
afd: The problem with this is that you are saying that a uniform is used just because some variable with the same name as it is declared.  You need to actually check that this variable is a uniform.  You can do this by checking that the base type of the variables declaration has the UNIFORM qualifier.

afd:
Having thought about this and discussed with @jiradeto, I suggest the following:

- Class RemoveRedundantUniformMetadataReductionOpportunities should not extend or implement anything

- It should have a static method, findOpportunities, with the right signature (the same one that all of the "opportunities" classes have)

- This static method should:

1. Make a list of the names of all uniforms in the pipeline info, "canBeRemoved"
2. Iterate over all shaders in the shader job
3. For each shader, use an anonymous subclass of StandardVisitor to search for declarations of uniforms, and remove the name of each such uniform from "canBeRemoved"
4. Finally, return a list of reduction opportunities, one for each of the remaining names.

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
```diff
@@ -1011,7 +1011,14 @@ private String getMD5(File shaderJobFile) throws IOException {
     byte[] computeData = isFile(computeShaderFile)
         ? readFileToByteArray(computeShaderFile)
         : new byte[0];
-    byte[] combinedData = new byte[vertexData.length + fragmentData.length + computeData.length];
+    // This metadata is required since in the case of
```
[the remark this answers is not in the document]
jiradeto: Done

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
jiradeto: Fixed

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunitiesTest.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/RemoveRedundantUniformMetadataReductionOpportunities.java
jiradeto: Thank you, fixed it.

PR #492, opened 2019-05-20
Add testIdentityNotNestedRemoved

paulthomson:
Add test for #491. This can be merged after #489, as this test might end up being redundant.

### DIFF
--- a/reducer/src/test/java/com/graphicsfuzz/reducer/util/SimplifyTest.java
+++ b/reducer/src/test/java/com/graphicsfuzz/reducer/util/SimplifyTest.java
@@ -127,14 +127,14 @@ public class SimplifyTest {
   @Test
   public void testIdentityNotNestedRemoved() throws Exception {
     final TranslationUnit tu = ParseHelper.parse("void main() {"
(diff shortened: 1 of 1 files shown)


reducer/src/test/java/com/graphicsfuzz/reducer/util/SimplifyTest.java
```diff
@@ -122,4 +122,24 @@ public void testMacroBlockRemoved() throws Exception {
     CompareAsts.assertEqualAsts(expected, simplifiedTu);
   }
 
+  // TODO(491): Enable once issue 491 is fixed.
+  @Ignore
+  @Test
+  public void testIdentityNotNestedRemoved() throws Exception {
+    final TranslationUnit tu = ParseHelper.parse("void main() {"
+        + " if(_GLF_IDENTITY(1, 1)) {}"
```
afd: I would prefer this to be _GLF_IDENTITY(true, true), since if(1) is not legal GLSL.

paulthomson: Done

PR #494, opened 2019-05-20
Update a few maven plug-in versions and fix some warnings

### LINKED ISSUE #495: Update checkstyle, jetty, and some other maven plugins

paulthomson:
Fix #495

### DIFF
--- a/checkstyle-config/src/main/resources/graphicsfuzz/checkstyle.xml
+++ b/checkstyle-config/src/main/resources/graphicsfuzz/checkstyle.xml
@@ -74,8 +74,8 @@ limitations under the License.
             <property name="tokens" value="LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_DO"/>
         </module>
         <module name="RightCurly">
-            <property name="id" value="RightCurlyAlone"/>
-            <property name="option" value="alone"/>
+            <property name="id" value="RightCurlySame"/>
+            <property name="option" value="same"/>
             <property name="tokens" value="CLASS_DEF, METHOD_DEF, CTOR_DEF, LITERAL_FOR, LITERAL_WHILE, STATIC_INIT, INSTANCE_INIT"/>
         </module>
         <module name="WhitespaceAround">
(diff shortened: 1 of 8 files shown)


parent-checkstyle/pom.xml
```diff
@@ -39,6 +39,10 @@ limitations under the License.
           <!-- declared in parent -->
         </dependencies>
         <configuration>
+          <sourceDirectories>
+            <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
+<!--            <sourceDirectory>${project.build.testSourceDirectory}</sourceDirectory>-->
```
afd: Is this to pave the way for us to run CheckStyle on the tests should we wish to do so?

checkstyle-config/src/main/resources/graphicsfuzz/checkstyle.xml
```diff
@@ -74,8 +74,8 @@ limitations under the License.
             <property name="tokens" value="LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_DO"/>
         </module>
         <module name="RightCurly">
-            <property name="id" value="RightCurlyAlone"/>
-            <property name="option" value="alone"/>
+            <property name="id" value="RightCurlySame"/>
+            <property name="option" value="same"/>
```
afd: I'm not sure what effect these changes have - are they in line with Google style?

parent-checkstyle/pom.xml
paulthomson: Yes

checkstyle-config/src/main/resources/graphicsfuzz/checkstyle.xml
paulthomson: Yes. There might even be more changes to the Google style since our snapshot, but this was enough to get our code passing.

paulthomson:
Indeed, fixing the warnings was not necessary, but I was planning to also increase the warning level of the Java compiler (which required fixing these issues), but then gave up as these additional flags apparently trigger a compiler bug!

parent-checkstyle/pom.xml
afd: Great.

PR #502, opened 2019-05-23
Compute shaders in WebUI

afd:
This PR adds some basic support for viewing compute shader results in the WebUI.  It does not yet allow reductions to be kicked off from there, but you can at least browse crashes and mismatches.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
@@ -1234,21 +1234,24 @@ public class ShaderJobFileOperations {
             // In the case that we do not have an exact match, store the output obtained by exact
             // diffing (as it may be useful to inspect).
             computeShaderComparisonWithReference.addProperty("exactdiff_output",
-                exactDiffResult.stdout.toString());
+                exactDiffResult.stderr.toString());
 
             // Now perform a fuzzy diff.
             final ExecResult fuzzyDiffResult =
-                new ExecHelper().exec(ExecHelper.RedirectType.TO_BUFFER,
-                    null, false,
-                    ToolPaths.getPythonDriversDir(),
-                    "inspect-compute-results",
+                new ExecHelper().exec(
+                    ExecHelper.RedirectType.TO_BUFFER,
+                    null,
+                    false,
+                    Paths.get(
+                        ToolPaths.getPythonDriversDir(),
+                        "inspect-compute-results").toString(),
                     "fuzzydiff",
                     referenceShaderResultFile.get().getAbsolutePath(),
                     shaderJobResultFile.getAbsolutePath());
             computeShaderComparisonWithReference.addProperty("fuzzy_match",
                 fuzzyDiffResult.res == 0);
             computeShaderComparisonWithReference.addProperty("fuzzydiff_output",
-                fuzzyDiffResult.stdout.toString());
+                fuzzyDiffResult.stderr.toString());
           }
           infoJson.add("comparision_with_reference", computeShaderComparisonWithReference);
         } catch (InterruptedException interruptedException) {
--- a/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
@@ -788,7 +788,7 @@ public class ShaderJobFileOperations {
   public void writeShaderResultToFile(
       ImageJobResult shaderResult,
       File shaderResultFile,
-      Optional<File> referenceShaderResultFile) throws IOException {
+      Optional<File> referenceShaderResultFile) throws InterruptedException, IOException {
 
     writeShaderResultToFileHelper(
         shaderResult,
@@ -1167,7 +1167,7 @@ public class ShaderJobFileOperations {
       ImageJobResult shaderResult,
       File shaderJobResultFile,
       ShaderJobFileOperations fileOps,
-      Optional<File> referenceShaderResultFile) throws IOException {
+      Optional<File> referenceShaderResultFile) throws InterruptedException, IOException {
 
     assertIsShaderJobResultFile(shaderJobResultFile);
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
```diff
@@ -1170,10 +1172,23 @@ protected static void writeShaderResultToFileHelper(
...
+          shaderResult.getLog());
+    }
 
+    // Special case: compute shader job.
     if (shaderResult.isSetComputeOutputs()) {
 
+      // In addition to a status and log, results for a compute shader job have:
+      // - an "outputs" property, which maps to a dictionary representing the results
+      //   that were obtained by running the shader;
+      // - (if reference result is present) a "comparision_with_reference" property
```
paulthomson: Typo: comparison

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
```diff
@@ -1186,21 +1201,80 @@ protected static void writeShaderResultToFileHelper(
...
+                    "fuzzydiff",
+                    referenceShaderResultFile.get().getAbsolutePath(),
+                    shaderJobResultFile.getAbsolutePath());
+            computeShaderComparisonWithReference.addProperty("fuzzy_match",
+                fuzzyDiffResult.res == 0);
+            computeShaderComparisonWithReference.addProperty("fuzzydiff_output",
+                fuzzyDiffResult.stderr.toString());
+          }
+          infoJson.add("comparison_with_reference", computeShaderComparisonWithReference);
+        } catch (InterruptedException interruptedException) {
```
paulthomson: I would just add the InterruptedException to the method signature, or at least rethrow it; I don't think it should happen unless the user tries to terminate the server. I also want to ban "printStackTrace" via a Python script or something, but I am lazy.

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
```diff
@@ -1186,21 +1201,80 @@ protected static void writeShaderResultToFileHelper(
...
+        //   fuzzy diffing.
+
+        final JsonObject computeShaderComparisonWithReference = new JsonObject();
+        try {
+
+          // Check whether the results exactly match those of the reference.
+          final ExecResult exactDiffResult =
+              new ExecHelper().exec(ExecHelper.RedirectType.TO_BUFFER,
+              null, false,
+                  Paths.get(ToolPaths.getPythonDriversDir(),"inspect-compute-results").toString(),
```
paulthomson: Does this work on Windows? If not, it might work if we just remove the python drivers path and assume that "inspect-compute-results" is on the PATH, which is not ideal, but making sure the python code always runs as expected from the command line was non-trivial, so I think it is OK to make use of the PATH.

server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
```diff
@@ -255,6 +287,28 @@ private static ImageDifferenceResultSet getImageDiffResult(JsonObject info) {
...
+    }
+
+    final JsonObject comparisonJson = info.get(comparisonWithReferenceKey).getAsJsonObject();
+
+    if (comparisonJson.has("exact_match") && comparisonJson.get("exact_match").getAsBoolean()) {
+      return ComputeDifferenceResult.IDENTICAL;
+    }
+
+    if (comparisonJson.has("fuzzy_match") && comparisonJson.get("fuzzy_match").getAsBoolean()) {
+      return ComputeDifferenceResult.IDENTICAL;
```
paulthomson: This looks wrong?

server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
```diff
@@ -567,31 +621,33 @@ private void worker(HttpServletRequest request, HttpServletResponse response)
 
       // TODO(360): Show results for compute shaders. For now, just indicate that some results
       // exists, and point to documentation.
-      if (shaderFamilyResult.shaderFamily.isComp) {
+      if (shaderFamilyResult.shaderFamily.isCompute) {
```
paulthomson: Maybe update the TODO?

server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
```diff
@@ -865,19 +921,24 @@ private void viewResult(HttpServletRequest request, HttpServletResponse response
     final String worker = path[3];
     final String shaderFamily = path[4];
     final String variant = path[5];
-    final String variantDir =
-        WebUiConstants.WORKER_DIR + "/" + worker + "/" + shaderFamily + "/";
-    final String variantFullPathNoExtension = variantDir + variant;
+    final Path variantDir =
+        Paths.get(WebUiConstants.WORKER_DIR, worker, shaderFamily);
+    final Path variantFullPathNoExtension = Paths.get(variantDir.toString(), variant);
```
paulthomson: Thanks for fixing these

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
afd: Fixed.

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
afd: I have added it to the method signature.  I also got rid of another printStackTrace from later in the method.

server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
afd: Good catch!  Fixed.

server/src/main/java/com/graphicsfuzz/server/webui/WebUi.java
afd: I have removed it.

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
afd: I think it should work as there is a .bat file in the drivers directory, but I will check.

afd:
No rush - let's check on Windows in due course.

common/src/main/java/com/graphicsfuzz/common/util/ShaderJobFileOperations.java
afd: I got around to checking this and you are correct: it gives an error.  If I add .bat on to the end of the command then things work.

We could add .bat if the server is running under Windows.  That would avoid the need to assume that the script is on the path.  Are there any downsides to detecting the OS and adding .bat if it is Windows?

paulthomson: If that works, then that sounds good enough. Maybe we should have a "runDriver" method for that.

afd: Done.

afd: ...and tested it on Linux and Windows.

PR #505, opened 2019-05-24
Fixed issue where attempts to make array access in donated code in bounds were being thwarted

### LINKED ISSUE #503: Out-of-bounds accesses are being generated in live code donation
In recent testing from @abelbriggs1 we have seen live code injections such as:

```
   {
    vec4 GLF_live1c = _GLF_FUZZED(vec4(7.9, -8.5, -3.7, 554.721));
    {
     vec4 _GLF_outVarBackup_GLF_color;
     _GLF_outVarBackup_GLF_color = _GLF_color;
     _GLF_color = (true ? vec4(-6.3, 595.143, -933.631, 357.018) : vec4(-4.9, -9175.5798, -261.894, 8.8));
     if(_GLF_WRAPPED_IF_TRUE(true))
      {
       _GLF_color = _GLF_outVarBackup_GLF_color;
      }
    }
    int GLF_live1i = _GLF_FUZZED(-759);
    if(_GLF_IDENTITY(GLF_live1c[GLF_live1i], clamp(GLF_live1c[GLF_live1i], GLF_live1c[GLF_live1i], _GLF_IDENTITY(GLF_live1c[GLF_live1i], (GLF_live1c[GLF_live1i]) - 0.0))) >= 1.0)
     {
      GLF_live1c[GLF_live1i] = GLF_live1c[GLF_live1i] * GLF_live1c[GLF_live1i];
     }
   }
```

The problem here is that indexing GLF_live1c using GLF_live1i is out-of-bounds.

At line 254 of DonateCodeTransformation.java, a method makeInjectedArrayAccessesInBounds is invoked which should make such array and vector accesses in bounds.

Either this method is not getting invoked on the right statements, or something is wrong with the logic for making access in bounds.

afd:
Fixes #503

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateCodeTransformation.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateCodeTransformation.java
@@ -33,7 +33,6 @@ import com.graphicsfuzz.common.ast.stmt.NullStmt;
 import com.graphicsfuzz.common.ast.stmt.Stmt;
 import com.graphicsfuzz.common.ast.type.ArrayType;
 import com.graphicsfuzz.common.ast.type.BasicType;
-import com.graphicsfuzz.common.ast.type.LayoutQualifier;
 import com.graphicsfuzz.common.ast.type.LayoutQualifierSequence;
 import com.graphicsfuzz.common.ast.type.QualifiedType;
 import com.graphicsfuzz.common.ast.type.StructDefinitionType;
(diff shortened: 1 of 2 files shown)


afd:
To give a bit more detail on this:

- After code is donated to a shader, functions from the donor are also donated in case the donated code calls them.

- In addition, prototypes for those functions are added at the top of the shader - for each function f being donated, the whole function is donated, an also an extra copy of the function's prototype.

- The function prototypes are extracted from the full functions, and thus need to be cloned so that they are truly distinct objects in the AST.

- To handle this, I had been lazy and cloned all declarations in the AST as part of the function prototype donation process.

- It was this cloning of everything that had led to the code to make access in bounds have no effect - the statements to be made in bounds no longer existed, as they had been cloned away.

- This PR fixes things by only cloning the function prototypes.

Issue #510, opened 2019-05-31
Add identity transformation to rewrite vector or matrix variable


afd:
Given a VariableIdentifierExpr a with matrix or vector type, we would like an identity transformation that turns a into:

type_constructor(a[0], ..., IDENTITY(a[random_index]), ..., a[n])

For example, if a has type mat3x3 then, if we chose random_index to be 2, we would turn a into:

mat3x3(a[0], a[1], IDENTITY(a[2])).

I suggest calling this IdentityRewriteComposite.

For it to inherit from AbstractIdentityTransformation it would be necessary to remove "final" from the "preconditionHolds" method of that class, and override "preconditonHolds" in IdentityRewriteComposite so that as well as checking that the precondition holds according to AbstractIdentityTransformation, we check that the expression is a VariableIdentifierExpression.

The motivation for limiting this to VariableIdentifierExpressions is to avoid blow-up.

afd:
Fixed via #512

PR #512, opened 2019-05-31
Add new transformation for rewriting nonscalars as their constructors

abelbriggs1:
Implements #510.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -44,6 +44,7 @@ import com.graphicsfuzz.util.Constants;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
@@ -78,6 +79,8 @@ public final class OpaqueExpressionGenerator {
     expressionIdentities.add(new IdentityMin());
     expressionIdentities.add(new IdentityMax());
     expressionIdentities.add(new IdentityClamp());
+
+    expressionIdentities.add(new IdentityRewriteComposite());
 
     if (shadingLanguageVersion.supportedMixNonfloatBool()) {
       expressionIdentities.add(new IdentityMixBvec());
@@ -530,7 +533,7 @@ public final class OpaqueExpressionGenerator {
     }
 
     @Override
-    public final boolean preconditionHolds(Expr expr, BasicType basicType) {
+    public boolean preconditionHolds(Expr expr, BasicType basicType) {
       if (!acceptableTypes.contains(basicType)) {
         return false;
       }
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+      // v -> (true ? vecX(..., identity(v[Y]), ...) : _)
+      // v -> (false ? _ : vecX(..., identity(v[Y]), ...))
+      // where v is a vector of size X, y is the random entry in v we want to apply
+      // identities to, and ... is the other entries in v that we don't change.
+      // Similarly for matrices.
+
+      assert BasicType.allVectorTypes().contains(type)
+          || BasicType.allMatrixTypes().contains(type);
+
+      final int numColumns =
```
afd: Even though it is clunky, I would favour "numElementsOrColumns" as a name here, because in the case of a vector we call them elements.

But actually, how about "numIndices"?  That would be nice as you later have a variable called "indexToApplyIdentities".

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+          (BasicType.allVectorTypes().contains(type)
+          ? type.getNumElements() : type.getNumColumns());
+      final int indexToApplyIdentities = generator.nextInt(numColumns);
+      final List<Expr> nonscalarEntryList = new ArrayList<>();
```
afd: ^ How about calling this "typeConstructorArguments"?  Because that's ultimately what they are going to be used for.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+    public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
+                      Fuzzer fuzzer) {
+      // v -> (true ? vecX(..., identity(v[Y]), ...) : _)
+      // v -> (false ? _ : vecX(..., identity(v[Y]), ...))
+      // where v is a vector of size X, y is the random entry in v we want to apply
+      // identities to, and ... is the other entries in v that we don't change.
+      // Similarly for matrices.
+
+      assert BasicType.allVectorTypes().contains(type)
+          || BasicType.allMatrixTypes().contains(type);
```
afd: Also assert that expr is a VariableIdentifierExpr - that would be a nice sanity check.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+
+      final int numColumns =
+          (BasicType.allVectorTypes().contains(type)
+          ? type.getNumElements() : type.getNumColumns());
+      final int indexToApplyIdentities = generator.nextInt(numColumns);
+      final List<Expr> nonscalarEntryList = new ArrayList<>();
+      for (int i = 0; i < numColumns; i++) {
+        // v + vec2(0.0)[0]; is not what we want to do, so we wrap expr into
```
afd: I had an original comment: Please rewrite comment to be a bit more descriptive.  E.g.: "If we are transforming an expression that already has the form 'v + vec2(0.0)', we would not want to generate an expression of the form 'v + vec2(0.0)[0]' to index into this; we would instead want it to look like '(v + vec2(0.0))[0]' ..."

But it then occurred to me that since we limit expr to being a VariableIdentifierExpr, we don't have this problem and do not need to add parentheses at all, so please simplify the code accordingly.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+        // parentheses for (v + vec2(0.0))[0]
```
afd: I would simplify this code to something like:

Expr argument = new ArrayIndexExpr(expr.clone(), new IntConstantExpr(...));
if (i == indexToApplyIdentities) {
  argument = applyIdentityFunction(argument, ...);
}
// And then add argument to the list of arguments.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+      // identities to, and ... is the other entries in v that we don't change.
+      // Similarly for matrices.
+
+      assert BasicType.allVectorTypes().contains(type)
+          || BasicType.allMatrixTypes().contains(type);
+
+      final int numColumns =
+          (BasicType.allVectorTypes().contains(type)
+          ? type.getNumElements() : type.getNumColumns());
+      final int indexToApplyIdentities = generator.nextInt(numColumns);
```
afd: ^ Maybe "indexToFurtherTransform"?

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -802,4 +804,60 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+    @Override
+    public boolean preconditionHolds(Expr expr, BasicType basicType) {
+      if (!super.preconditionHolds(expr, basicType)) {
+        return false;
+      }
+      if (!(expr instanceof VariableIdentifierExpr)) {
+        // limit blowup of identities.
+        return false;
+      }
+      return SideEffectChecker.isSideEffectFree(expr, shadingLanguageVersion);
```
afd: ^ super.preconditionHolds(...) does the side-effect checking.  So it is redundant and confusing to do it again here.

I would thus change the whole method body to:

return super.preconditionHolds(expr, basicType) && expr instanceof VariableIdentifierExpr;

PR #518, opened 2019-06-04
Add support for built-in GLES 3.20 Integer Functions

abelbriggs1:
Implements #467. 

OpenGL versions including and above 3.1ES and 4.0 support integer functions. The function signatures do not change between versions, nor were any functions added after 3.1ES, so I went ahead and added to the ShadingLanguageVersion interface instead of manually checking all the versions in TyperHelper.

Regarding the change in TyperTest, the test was not set up to handle void-returning functions, producing statements like:

```
uint testwhatever()
{
   return imul(u0, u1, u2, u3);
}
```

when imul() is a void function. This would cause a validation error, so I added a void type check.


abelbriggs1:
I needed to delete and repush the branch due to a rebase issue, so I'll make another PR.

PR #521, opened 2019-06-05
Add built-in support for GLSL Integer Functions

abelbriggs1:
Implements what can currently be implemented of #467. There are some functions that cannot be supported yet due to issues #522 and #524.

OpenGL versions including and above 3.1ES and 4.0 support integer functions. The function signatures do not change between versions, nor were any functions added after 3.1ES, so I went ahead and added to the ShadingLanguageVersion interface instead of manually checking all the versions in TyperHelper.

Regarding the change in TyperTest, the test was not set up to handle void-returning functions, producing statements like:
```
uint testwhatever()
{
   return imul(u0, u1, u2, u3);
}
```
when imul() is a void function. This would cause a validation error, so I added a void type check.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
@@ -177,6 +177,8 @@ public interface ShadingLanguageVersion {
 
   boolean supportedIntBitsToFloat();
 
+  boolean supportedIntegerFunctions();
+
   boolean supportedInverse();
 
   boolean supportedIsinf();
--- a/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
@@ -178,9 +178,9 @@ public interface ShadingLanguageVersion {
   boolean supportedIntBitsToFloat();
 
   /**
-   * Integer Functions are a set of built-in functions (think sqrt, max, min, etc.) that
-   * allow manipulation of integers and their vectors in ways difficult or impossible with normal
-   * GLSL syntax - for example, summing two unsigned integers where the result causes an overflow.
+   * Integer Functions are a set of built-in functions that allow manipulation of integers and
+   * their corresponding vectors in ways difficult or impossible with normal GLSL syntax - for
+   * example, summing two unsigned integers where the result causes an overflow.
    * GLSL versions 4.0+ and ESSL versions 3.1+ support these functions.
    *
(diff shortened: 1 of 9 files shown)


abelbriggs1:
(This PR will require #525 to be merged before it passes tests - the test failures here are related to that PR.)

ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -177,6 +177,8 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
 
   boolean supportedIntBitsToFloat();
```
afd: Can you add a JavaDoc comment for this new method, explaining what 'integer functions' means with regards to the spec?

ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
```diff
@@ -840,13 +841,81 @@ public static Type resolveTypeOfMul(Type lhsType, Type rhsType) {
...
+
     // 8.13: Fragment Processing Functions (only available in fragment shaders)
 
     // 8.14: Noise Functions - deprecated, so we do not consider them
 
     return builtinsForVersion;
   }
 
+  private static void getBuiltinsForGlslVersionInteger(
+      Map<String, List<FunctionPrototype>> builtinsForVersion) {
```
afd: Change TODO to TODO(GitHub issue id)
Remove TODO from second line, but indent enough so that IDEA still highlights the line in yellow.

ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -177,6 +177,16 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
...
 
+  /**
+   * Integer Functions are a set of built-in functions (think sqrt, max, min, etc.) that
```
afd: I think your examples are wrong.

In the GLSL ES 3.20 specification, the integer functions are things like uaddCarry.  Can you check and re-work?  sqrt, max, min, etc., work on various types - not just integers, and they are not in the "Integer Functions" section of the spec.

abelbriggs1: Sorry - the way I meant that was as examples of built-in functions in general, not as examples of integer functions specifically. The purpose was to explain what a built-in would be considered as to someone coming from a different language (Java, C++, etc.), but I suppose it's not really needed in this context.

abelbriggs1:
I've readded the functions that require out parameters, since I put up PR #533 that fixes issues #522 and #524. This PR will need that one merged before it passes tests.

Issue #517, opened 2019-06-04
Enhance identity transformation to rewrite matrix variable


afd:
Currently, thanks to #512, we can rewrite e.g. a mat2x2, m, to e.g. mat2x2(m[0], IDENTITY(m[1])).

It would be nice if we also gave the fuzzer the option to rewrite the matrix as e.g. mat2x2(m[0][0], IDENTITY(m[0][1]), m[1][0], m[1][1]).  (I may not have got this right - can never remember whether rows or columns come first in GLSL!).

abelbriggs1:
Fixed by #527.

PR #527, opened 2019-06-06
Enhance matrix constructor identity to potentially use all elements of matrix

abelbriggs1:
Implements #517.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
@@ -457,6 +457,12 @@ public class BasicType extends BuiltinType {
     return 4;
   }
 
+  /**
+   * Finds the number of rows in a matrix type. Will cause an assertion failure if
+   * used with a non-matrix type.
+   *
+   * @return the number of rows that the matrix type has.
+   */
   public int getNumRows() {
     assert allMatrixTypes().contains(this);
     if (Arrays.asList(BasicType.MAT2X2, BasicType.MAT3X2, BasicType.MAT4X2).contains(this)) {
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
@@ -456,4 +456,16 @@ public class BasicType extends BuiltinType {
     assert Arrays.asList(BasicType.MAT4X2, BasicType.MAT4X3, BasicType.MAT4X4).contains(this);
     return 4;
   }
+
+  public int getNumRows() {
+    assert allMatrixTypes().contains(this);
+    if (Arrays.asList(BasicType.MAT2X2, BasicType.MAT3X2, BasicType.MAT4X2).contains(this)) {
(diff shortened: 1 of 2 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -823,19 +823,40 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
       final List<Expr> typeConstructorArguments = new ArrayList<>();
-      for (int i = 0; i < numIndices; i++) {
-        Expr argument = new ArrayIndexExpr(expr.clone(), new IntConstantExpr(String.valueOf(i)));
-        if (i == indexToFurtherTransform) {
-          argument = applyIdentityFunction(argument,
-              (type.isVector() ? type.getElementType() : type.getColumnType()),
-              constContext, depth, fuzzer);
+
+      if (type.isMatrix() && generator.nextBoolean()) {
+        // further break down matrix constructor into single components, to increase diversity
```
afd: Elaborate the last sentence of the comment more. "For example, a mat2x2 m might turn into: ..."

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -823,19 +823,40 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+        // GLSL is column-major. mat2x2(m[0][0], m[0][1], IDENTITY(m[1][0]), m[1][1])
+        final int rowToFurtherTransform = generator.nextInt(type.getNumRows());
+        for (int i = 0; i < numColumns; i++) {
+          for (int j = 0; j < type.getNumRows(); j++) {
```
afd: Start comment with capital, and make the comment less note-like; e.g. "The inner ArrayIndexExpr is the column (first) index, and the outer one is the row (second) index."

ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
```diff
@@ -456,4 +456,22 @@ public int getNumColumns() {
     assert Arrays.asList(BasicType.MAT4X2, BasicType.MAT4X3, BasicType.MAT4X4).contains(this);
     return 4;
   }
+
+  /**
+   * Finds the number of rows in a matrix type. Will cause an assertion failure if
```
afd: Actually how about we do this properly - make it throw an unsupported operation exception (with a message string) instead of asserting, and use a @throws clause in the JavaDoc to document that this can happen?

In general don't say "Will cause an assertion failure if" because assertions are not always enabled, so this depends on what mode you run java in (whether -ea is passed or not).

"Can only be invoked on a matrix type, and finds the number of rows of the matrix."

afd: But in fact I suggest that you do this as a separate PR, and do it for all the functions in the class that have that form.

Issue #480, opened 2019-05-16
reducer: replace VariablesDeclaration with expression


paulthomson:
When not preserving semantics, the reducer should try:

```glsl
void main()
{
 bool doSwap = checkSwap();
}
```

```glsl
void main()
{
 checkSwap();
}
```

paulthomson:
Could do this?

```glsl
bool doSwap = checkSwap();
->
bool doSwap; doSwap = checkSwap();
```

paulthomson:
int a = 42, b, c = foo(), d;
->
int a, b, c, d;
a = 42;
c = foo();

paulthomson:
This can't be semantics-preserving, becuase e.g. we can do:

int a = 12, b = a;

Or:

int a = foo(), b = bar(); // foo and bar have side effects; the effects of bar depend on what foo did

afd:
Fixed via #529 and #576.

Issue #520, opened 2019-06-05
Decide whether to support initializer lists, and revisit ScalarInitializer name


afd:
OpenGL (but not OpenGL ES) shading languages allow initializer lists for arrays, e.g.:

int A[3] = { 1, 2, 3 };

GraphicsFuzz has a placeholder class, ArrayInitializer, for this, but it currently doesn't do anything and the GraphicsFuzz AST builder throws an exception if it encounters an initializer list.  However, regular array initializers are supported, e.g.:

int A[3] = int[3](1, 2, 3);

Confusingly, anything that is not an initializer list is treated as a ScalarInitializer in the tool, so an array initializer like the above example falls into the ScalarInitializer category.

I propose that we:

(1) Make the runtime exception that gets thrown when an initializer list is encountered more explicit that this case is not supported;

(2) Get rid of the ArrayInitializer class;

(3) Make what is currently the ScalarInitializer class just be the Initializer class.

afd:
Fixed via #601

PR #529, opened 2019-06-07
Implementation: Reducer Opportunity to replace variable declaration with an expression

jiradeto:
Related issue #480:

Implemented new reducer that will replace variable declarations with the binary expressions.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
@@ -92,7 +92,8 @@ public class ReductionDriver {
         IReductionOpportunityFinder.inlineFunctionFinder(),
         IReductionOpportunityFinder.unusedParamFinder(),
         IReductionOpportunityFinder.foldConstantFinder(),
-        IReductionOpportunityFinder.redundantUniformMetadataFinder())) {
+        IReductionOpportunityFinder.redundantUniformMetadataFinder(),
+        IReductionOpportunityFinder.variableDeclToExprFinder())) {
       cleanupPasses.add(new SystematicReductionPass(context,
           verbose,
           finder));
@@ -119,8 +120,7 @@ public class ReductionDriver {
         IReductionOpportunityFinder.liveFragColorWriteFinder(),
         IReductionOpportunityFinder.functionFinder(),
         IReductionOpportunityFinder.variableDeclFinder(),
-        IReductionOpportunityFinder.globalVariablesDeclarationFinder(),
-        IReductionOpportunityFinder.variableDeclToExprFinder())) {
+        IReductionOpportunityFinder.globalVariablesDeclarationFinder())) {
       final SystematicReductionPass pass = new SystematicReductionPass(context,
           verbose,
           finder);
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
@@ -119,7 +119,8 @@ public class ReductionDriver {
         IReductionOpportunityFinder.liveFragColorWriteFinder(),
         IReductionOpportunityFinder.functionFinder(),
         IReductionOpportunityFinder.variableDeclFinder(),
-        IReductionOpportunityFinder.globalVariablesDeclarationFinder())) {
+        IReductionOpportunityFinder.globalVariablesDeclarationFinder(),
+        IReductionOpportunityFinder.variableDeclToExprFinder())) {
       final SystematicReductionPass pass = new SystematicReductionPass(context,
           verbose,
           finder);
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
@@ -17,6 +17,8 @@
 package com.graphicsfuzz.reducer.reductionopportunities;
 
 import com.graphicsfuzz.common.ast.TranslationUnit;
+import com.graphicsfuzz.common.ast.decl.VariableDeclInfo;
+import com.graphicsfuzz.common.ast.stmt.DeclarationStmt;
 import com.graphicsfuzz.common.transformreduce.ShaderJob;
 import com.graphicsfuzz.common.util.ListConcat;
 import java.util.Arrays;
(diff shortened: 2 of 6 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
```diff
@@ -119,7 +119,8 @@ public ReductionDriver(ReducerContext context,
         IReductionOpportunityFinder.liveFragColorWriteFinder(),
         IReductionOpportunityFinder.functionFinder(),
         IReductionOpportunityFinder.variableDeclFinder(),
-        IReductionOpportunityFinder.globalVariablesDeclarationFinder())) {
+        IReductionOpportunityFinder.globalVariablesDeclarationFinder(),
```
afd: I regard this as a cleanup pass rather than a core pass; can you move it up so that it's only added to cleanup passes?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
```diff
@@ -53,4 +55,28 @@ public VariableDeclToExprReductionOpportunities(TranslationUnit tu, ReducerConte
     finder.visit(tu);
     return finder.getOpportunities();
   }
+
```
afd: Could you add a comment to the top of this class briefly explaining, with a couple of examples, the kinds of opportunity that this pass will look for?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
```diff
@@ -16,20 +16,49 @@
...
+import com.graphicsfuzz.common.ast.expr.BinaryExpr;
+import com.graphicsfuzz.common.ast.expr.VariableIdentifierExpr;
+import com.graphicsfuzz.common.ast.stmt.BlockStmt;
+import com.graphicsfuzz.common.ast.stmt.DeclarationStmt;
+import com.graphicsfuzz.common.ast.stmt.ExprStmt;
 import com.graphicsfuzz.common.ast.visitors.VisitationDepth;
 
 public class VariableDeclToExprReductionOpportunity extends AbstractReductionOpportunity {
 
-  VariableDeclToExprReductionOpportunity(VisitationDepth depth) {
```
afd: Can you add some comments about these fields, explaining what they represent in terms of the reduction opportunity and what they represent?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
```diff
@@ -53,4 +55,28 @@ public VariableDeclToExprReductionOpportunities(TranslationUnit tu, ReducerConte
...
+      // Replacing variable declaration with a new binary expression might have a side-effect,
+      // do not consider these reduction opportunities if we are not reducing everywhere.
+      return;
+    }
+    final List<VariableDeclInfo> declInfos =
+        declarationStmt.getVariablesDeclaration().getDeclInfos();
+    // We iterate backwards so that when applying reduction the new expression is inserted at the
+    // correct order with respect to its original order in the variable declaration info list.
+    for (int i = declInfos.size() - 1; i >= 0; i--) {
+      final VariableDeclInfo variableDeclInfo = declInfos.get(i);
```
afd: Also check that the initializer is a ScalarInitializer (as that's a condition we require later).

It might be that we get rid of the notion of ScalarInitializer due to #520, but still we should assume for now that there are multiple kinds of initializer and that we are only supporting one.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
```diff
@@ -53,4 +55,28 @@ public VariableDeclToExprReductionOpportunities(TranslationUnit tu, ReducerConte
...
+      if (variableDeclInfo.hasInitializer()) {
+        addOpportunity(new VariableDeclToExprReductionOpportunity(
+            variableDeclInfo,
```
afd: Because the current class extends ScopeTreeBuilder, you have access to the "currentBlock" method which gets the block that visitation is inside.  I suggest you get rid of parentMap and call this method to get the block.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
```diff
@@ -16,20 +16,49 @@
...
+  private final VariableDeclInfo variableDeclInfo;
+  private final DeclarationStmt declarationStmt;
```
afd: With respect to my above comment, replace "parent" with "private final Block enclosingBlock"

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
```diff
@@ -16,20 +16,49 @@
...
+        ((ScalarInitializer) variableDeclInfo.getInitializer()).getExpr(),
+        BinOp.ASSIGN
+    );
+    ((BlockStmt) parent).insertAfter(declarationStmt, new ExprStmt(binaryExpr));
+    variableDeclInfo.setInitializer(null);
   }
 
   @Override
   public boolean preconditionHolds() {
-    return false;
```
afd: If parent becomes enclosingBlock, the instanceof check can go.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
jiradeto: Nice! thank you for guiding.

reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
jiradeto: Updated.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunities.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
jiradeto: Fixed

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/VariableDeclToExprReductionOpportunity.java
jiradeto: Updated

PR #532, opened 2019-06-07
Refactor BasicType with documentation and exception improvements

### LINKED ISSUE #531: BasicType asserts instead of throwing exceptions for unsupported operations
```
public int getNumRows() {
    assert allMatrixTypes().contains(this);
    if (Arrays.asList(BasicType.MAT2X2, BasicType.MAT3X2, BasicType.MAT4X2).contains(this)) {
      return 2;
    }
    if (Arrays.asList(BasicType.MAT2X3, BasicType.MAT3X3, BasicType.MAT4X3).contains(this)) {
      return 3;
    }
    assert Arrays.asList(BasicType.MAT2X4, BasicType.MAT3X4, BasicType.MAT4X4).contains(this);
    return 4;
  }
```

Functions like these should probably throw UnsupportedOperationException instead of asserting.

abelbriggs1:
Fixes #531. 

This PR:

 - makes a distinction between usage of RuntimeException and UnsupportedOperationException - RuntimeException is still used in functions like toString(), where the type could possibly be in an invalid state. RuntimeException has been replaced by UnsupportedOperationException when the user can potentially (and realistically) misuse a function (sending invalid arguments or using an unintended type). Additionally, UnsupportedOperationException is thrown instead of asserting when an error state is realistically possible.

 - adds Javadoc comments to functions that throw UnsupportedOperationException.

 - replaces usage of static allTypes().contains(type) calls with the corresponding isType() call when the current scope is not static.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
@@ -344,26 +344,44 @@ public class BasicType extends BuiltinType {
           BVEC4));
   }
 
+  /**
+   * Static version of getNumColumns(). Finds the number of columns in a matrix. Can only be
+   * invoked on a matrix type.
+   *
+   * @return the number of columns that the matrix type has.
+   * @throws UnsupportedOperationException if the type is not a matrix.
+   */
   public static int numColumns(BasicType basicType) {
-    assert allMatrixTypes().contains(basicType);
+    if (!allMatrixTypes().contains(basicType)) {
+      throw new UnsupportedOperationException(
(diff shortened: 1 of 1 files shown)


ast/src/main/java/com/graphicsfuzz/common/ast/type/BasicType.java
```diff
@@ -344,26 +344,44 @@ public boolean isBoolean() {
           BVEC4));
   }
 
+  /**
```
afd: Hmm, not sure why we have this static method at all.  It looks like numColumns(mat) is the same as mat.getNumColumns().

Can you open an issue for us to delete the static method and change all code that used it to use the instance version?

Issue #524, opened 2019-06-05
FunctionCallExprTemplate doesn't support lvalue arguments


abelbriggs1:
```
  @Override
  public boolean requiresLValueForArgument(int index) {
    assert index >= 0 && index < getNumArguments();
    return false;
  }
```

The fuzzer simply assumes that built-in arguments don't require lvalues - this is a problem if we want to use a function that takes, for example, an `out` parameter.

abelbriggs1:
Fixed by #533.

PR #533, opened 2019-06-07
Check for side effects with lvalue parameters of built-in functions

### LINKED ISSUE #522: SideEffectChecker should not assume all built-ins are side-effect free
Currently, the SideEffectChecker assumes all built in GLSL functions are free of side effects that could affect other areas of the shader program, like changing global variables. With the advent of additional built-in functions like integer functions, which have `out` parameters, this is no longer always the case.

abelbriggs1:
Fixes #522 and #524. I made sure to test this outside of the IDE by readding built-ins (specifically the integer ones) with out parameters and generating hundreds of variants, then grepping for the offending functions to see if they were being used properly.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
@@ -17,6 +17,8 @@
 package com.graphicsfuzz.common.util;
 
 import com.graphicsfuzz.common.ast.IAstNode;
+import com.graphicsfuzz.common.ast.decl.FunctionPrototype;
+import com.graphicsfuzz.common.ast.decl.ParameterDecl;
 import com.graphicsfuzz.common.ast.expr.BinaryExpr;
 import com.graphicsfuzz.common.ast.expr.Expr;
 import com.graphicsfuzz.common.ast.expr.FunctionCallExpr;
@@ -28,6 +30,7 @@ import com.graphicsfuzz.common.ast.stmt.DiscardStmt;
 import com.graphicsfuzz.common.ast.stmt.ExprCaseLabel;
 import com.graphicsfuzz.common.ast.stmt.ReturnStmt;
 import com.graphicsfuzz.common.ast.stmt.Stmt;
+import com.graphicsfuzz.common.ast.type.TypeQualifier;
 import com.graphicsfuzz.common.ast.visitors.CheckPredicateVisitor;
 import com.graphicsfuzz.common.glslversion.ShadingLanguageVersion;
 import com.graphicsfuzz.common.typing.TyperHelper;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
```diff
@@ -45,6 +48,17 @@ public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
           // Assume that any call to a non-builtin might have a side-effect.
           predicateHolds();
         }
+        for (FunctionPrototype p :
+            TyperHelper.getBuiltins(shadingLanguageVersion).get(functionCallExpr.getCallee())) {
+          // We check each argument of the built-in's prototypes to see if they require lvalues -
+          // if so, they can cause side effects.
```
afd: Also add a note to say that we could try to be more precise by resolving exactly which overload of the builtin is being invoked.

common/src/test/java/com/graphicsfuzz/common/util/SideEffectCheckerTest.java
```diff
@@ -62,4 +65,12 @@ public void testCountingLoopHasSideEffects() throws Exception {
         ShadingLanguageVersion.ESSL_310));
   }
 
+  // TODO: Unignore when #521 is ready to merge.
```
afd: Please change to TODO(521): Unignore when #521 is ready to merge.

common/src/test/java/com/graphicsfuzz/common/util/SideEffectCheckerTest.java
```diff
@@ -62,4 +65,12 @@ public void testCountingLoopHasSideEffects() throws Exception {
...
+  @Ignore
+  @Test
```
afd: I would prefer this test if the expression was a more complete use of uaddCarry.

I suggest you write a simple shader with a main function that invokes uaddCarry in a few different ways, passing valid arguments each time and trying a few of the overloads of this function (if it has them; I think it does).

Then apply a CheckPredicateVisitor to the parsed translation unit for that function, and for each FunctionCallExpr visited (a) assert that the name of the called function is uaddCarry, and (b) that the expression is side-effect free.

common/src/test/java/com/graphicsfuzz/common/util/SideEffectCheckerTest.java
```diff
@@ -62,4 +71,30 @@ public void testCountingLoopHasSideEffects() throws Exception {
...
+        + "   uaddCarry(uint(0), uint(0), out1);"
+        + "   uaddCarry(uvec2(0, 0), uvec2(0, 0), out2);"
+        + "   uaddCarry(uvec3(0, 0, 0), uvec3(0, 0, 0), out3);"
+        + "   uaddCarry(uvec4(0, 0, 0, 0), uvec4(0, 0, 0, 0), out4);"
+        + "}";
+
+    final TranslationUnit tu = ParseHelper.parse(shader, ShaderKind.FRAGMENT);
+    tu.setShadingLanguageVersion(ShadingLanguageVersion.ESSL_310);
+
+    assertFalse(new CheckPredicateVisitor() {
```
afd: I like what you've done here, asserting regarding the name and side-effects each time a call using the built-in is reached.  Now that I see it I realise there's no point using a CheckPredicate visitor.  Please remove the outer assert and change it to:

new StandardVisitor() {
  // what you already have
}.visit(tu);

PR #539, opened 2019-06-11
Add AmberScript generation for image and compute, and relevant runspv tests

hevrard:
The AmberScript generated is usable, but it still misses features like
embedding the original GLSL as comment.


hevrard:
Tests and compute shaders are missing, I was slowed down by test issues.

This should be mergeable now, if you don't want to wait for me to add compute shader and tests.

hevrard:
This PR has been completely reworked, it now implements support for AmberScript format on both image and compute. Additional runspv tests have been added.

python/src/main/python/drivers/runspv.py
```diff
@@ -925,13 +958,159 @@ def amberscriptify_image(
...
+    result = ''
+    with open(uniform_json, 'r') as f:
+        j = json.load(f)
+    for name, entry in j.items():
+
+        if name == '$compute':
+            continue
+
+        func = entry['func']
+        if func not in UNIFORM_TYPE.keys():
```
afd: We should probably change the following to "raise"; @paulthomson what do you think?

python/src/main/python/drivers/runspv.py
```diff
@@ -925,13 +958,159 @@ def amberscriptify_image(
...
+    else:
+        result += 'SHADER vertex gfz_vert PASSTHROUGH\n\n'
+
+    result += 'SHADER fragment gfz_frag SPIRV-ASM\n'
+    result += spv_get_disassembly(frag)
+    result += 'END\n\n'
+
+    # This buffer MUST be named framebuffer to be able to retrieve the image
+    # Format MUST be B8G8R8A8_UNORM (other options may become available once
+    # Amber supports more format for image extraction)
```
afd: ^ format -> formats

PR #544, opened 2019-06-11
Add graphicsfuzz-tool for running an arbitrary Java class

paulthomson:


python/src/main/python/drivers/graphicsfuzz-tool.bat
```diff
@@ -0,0 +1,29 @@
+@echo off
```
afd: License header missing?

PR #547, opened 2019-06-11
Handle SSBO with several types

hevrard:
Amber does not support more than one type. Promote bools when
feasible.


afd:
Let's close this and adapt our compute shader samples to accept Amber's limitations for now.

Issue #548, opened 2019-06-11
Write new identities with bitwise operations


abelbriggs1:
```
    // TODO: there are many more identities that we can easily play with here, e.g. bitwise and 1
    // for integer types
```
in OpaqueExpressionGenerator, line 66.

There are currently no functions in OpaqueExpressionGenerator that make use of bitwise functions, but there's a lot of operators to play with.

Some examples of identities might include:

Using one's complement operators, such that:
   int i -> ~(~i)

Using the OR operator to OR an integer with itself, such that:
   int i -> (i | i)

abelbriggs1:
Implemented with #553.

PR #553, opened 2019-06-12
Add new identities for integer types that use bitwise operations

abelbriggs1:
Implements #548.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -725,27 +725,28 @@ public final class OpaqueExpressionGenerator {
 
   /**
    * Identity transformation for integer types (both unsigned and signed, and their vectors) that
-   * ORs an integer with itself, producing the same integer as output. When performed, transforms an
-   * expression, e, such that:
+   * ORs an integer with itself, producing the same integer as output. This identity requires
+   * expressions to be side effect free because the same mutated expression is evaluated twice.
+   * When performed, transforms an expression, e, such that:
    *    e -> (e) | (e).
    */
   private class IdentityBitwiseOr extends AbstractIdentityTransformation {
     private IdentityBitwiseOr() {
-      super(BasicType.allIntegerTypes(), false);
+      super(BasicType.allIntegerTypes(), true);
     }
 
     @Override
     public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
                       Fuzzer fuzzer) {
       assert BasicType.allIntegerTypes().contains(type);
-      // We use parentheses inside of the binary expression to prevent issues with order of
-      // operations - for example, d = d * 2u | d = d * 2u will cause issues.
+      // We use parentheses to prevent issues with order of operations in ternary expressions.
       return identityConstructor(
           expr,
-          new BinaryExpr(
-              new ParenExpr(applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer)),
-              new ParenExpr(applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer)),
-              BinOp.BOR));
+          new ParenExpr(
+              new BinaryExpr(
+                  applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer),
+                  applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer),
+                  BinOp.BOR)));
     }
   }
 
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -749,10 +749,16 @@ public final class OpaqueExpressionGenerator {
                   BinOp.BOR));
     }
   }
-
+  
+  /**
+   * Identity transformation for integer types (both unsigned and signed, and their vectors) that
+   * ORs an integer with zero, producing the same integer as output.
+   * When performed, transforms an expression, e, such that:
+   *    e -> (e) | (opaque 0).
+   */
   private class IdentityBitwiseOrZero extends AbstractIdentityTransformation {
     private IdentityBitwiseOrZero() {
-      super(BasicType.allIntegerTypes(), true);
+      super(BasicType.allIntegerTypes(), false);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -691,6 +697,105 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+  }
+
+  /**
+   * Identity transformation for integer types (both unsigned and signed, and their vectors) that
+   * ORs an integer with itself, producing the same integer as output. When performed, transforms an
+   * expression, e, such that:
+   *    e -> (e) | (e).
+   */
+  private class IdentityBitwiseOr extends AbstractIdentityTransformation {
+    private IdentityBitwiseOr() {
```
afd: The second argument to "super" should be "true" here, since we're going to end up evaluating the mutated expression twice.  That means that it (in general) needs to be side-effect free.

afd: It would be worth adding a note too that effect in the comment for the private class.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -691,6 +697,105 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+      super(BasicType.allIntegerTypes(), false);
+    }
+
+    @Override
+    public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
+                      Fuzzer fuzzer) {
+      assert BasicType.allIntegerTypes().contains(type);
```
afd: Thanks for the example in the comment, but can you make it more explicit as to what is being transformed, e.g. "for example transforming ... to ... | ... would not preserve semantics."

(The use of the assignment operator in the example looks a bit odd to me - this is the kind of thing that would not be side-effect free and that therefore we shouldn't do, as d would get assigned to twice not once.)

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -63,18 +63,24 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    // TODO: there are many more identities that we can easily play with here
     expressionIdentities.add(new IdentityAddSubZero());
     expressionIdentities.add(new IdentityMulDivOne());
     expressionIdentities.add(new IdentityAndTrue());
     expressionIdentities.add(new IdentityOrFalse());
-    expressionIdentities.add(new IdentityNotNot());
+    expressionIdentities.add(new IdentityLogicalNotNot());
     expressionIdentities.add(new IdentityTernary());
 
+    if (shadingLanguageVersion.supportedBitwiseOperations()) {
```
afd: These look good!

How about also adding:

- IdentityBitwiseOrZero // e -> e | 0

And possibly change "IdentityBitwiseOr" to "IdentityBitwiseOrSelf" to help disambiguate the two.

I did also wonder about adding:

- IdentityBitwiseAndOnes // e -> e & 0xFFFFFFFF, but actually it's not clear to me whether 0xFFFFFFFF is always well-defined in GLSL due to the limited range of integers in some contexts.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Sorry for the ambiguity - the reason I put that as a specific example is because that is the exact expression I ran into issues with when running tests for that identity. The above fix, to ensure the expression must be side effect free, should fix this issue. I didn't really think about the fact that the assignment would have side effects, so I just patched that case instead of just preventing side effects for all expressions. Whoops.

abelbriggs1: Additionally, upon retesting without parentheses, I can't see from the expressions what exactly is going wrong with the failing tests. The one thing they have in common is that they're dealing with bitwise OR in ternary expressions.

I've cut down the comment (since I can't even find a good concrete example) to give the basic idea, and I was able to move the parentheses to outside the OR expression (instead of surrounding the two mutated expressions) without issue.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Done. Also see my comments in the below conversation about the comment, as they're related somewhat to this.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

abelbriggs1:
I didn't know there was a function that could do that, thanks for pointing it out. I've refactored to use applyBinaryIdentityFunction whenever possible.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -691,6 +698,124 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+    private IdentityBitwiseOrZero() {
+      super(BasicType.allIntegerTypes(), false);
+    }
+
+    @Override
+    public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
+                      Fuzzer fuzzer) {
+      assert BasicType.allIntegerTypes().contains(type);
+      return applyBinaryIdentityFunction(
+          expr.clone(),
```
afd: From studying IdentityAddSubZero, I don't think you need to put the opaque zero in a ParenExpr (i.e. I believe sufficient parentheses should come with the opaque zero).

Can you check the various places you add parentheses to see whether they are required?

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -691,6 +698,124 @@ public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
...
+              new ParenExpr(applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer)),
+              new ParenExpr(applyIdentityFunction(expr.clone(), type, constContext, depth, fuzzer)),
+                  BinOp.BOR));
+    }
+  }
+  
+  /**
+   * Identity transformation for integer types (both unsigned and signed, and their vectors) that
+   * ORs an integer with zero, producing the same integer as output.
+   * When performed, transforms an expression, e, such that:
```
afd: Update the comment to reflect the fact that the zero can be on either side, and similar for other relevant transformations.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: You're correct - I just tested them and they aren't required. Thanks for noticing that.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

PR #554, opened 2019-06-12
Add support for angle and trigonometric built-in functions

jiradeto:
Related issue: #460

The implementation of the new set of the built-in function for Angle and Trigonometric Functions. Following are the GLSL supported version for the new built-in functions.

General trigonometric functions (sin, cos, tan, etc):
- GLSL 1.1+
- GLSL ES 1.0+

Hyperbolic trigonometric functions (sinh, cosh, asinh, etc):
- GLSL 1.3+
- GLSL ES 3.0+

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
@@ -285,7 +285,7 @@ public final class TyperHelper {
   }
 
   public static Map<String, List<FunctionPrototype>> getBuiltins(ShadingLanguageVersion
-                                                                     shadingLanguageVersion) {
+      shadingLanguageVersion) {
     if (!builtins.containsKey(shadingLanguageVersion)) {
       builtins.putIfAbsent(shadingLanguageVersion,
           getBuiltinsForGlslVersion(shadingLanguageVersion));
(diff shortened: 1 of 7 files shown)


ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
```diff
@@ -297,49 +297,10 @@ public static Type resolveTypeOfMul(Type lhsType, Type rhsType) {
...
-      }
-    }
-    {
-      final String name = "atan";
-      for (Type t : genType()) {
-        addBuiltin(builtinsForVersion, name, t, t, t);
-      }
-    }
+    // 8.1: Angle and Trigonometric Functions
+
```
afd: @abelbriggs1 has taken the approach of checking here whether the relevant functions are supported; you have taken the approach of checking inside the "getBuiltinsForGlslVersionKINDOFFUCTIONS(...)" method.

It's a minor point but it would be good to be symmetric here.

I guess you did it inside because there a couple of different kinds of functions that came in different GLSL versions?  That seems like a good reason to do the checks inside.

@abelbriggs1 in that case perhaps in this PR @jiradeto should move the checks for your builtins inside the relevant "getBuiltinsForGlslVersionKINDOFFUNCTIONS(...)" method - what do you think?

abelbriggs1: That's probably a good idea. I was originally leaning towards any all-encompassing checks being outside and any subsets being inside the function (see integer functions), but it'll look rather unsightly by the end. It'd be better for each built-in adding function to handle the support checks.

afd: Thanks - can you do that please @jiradeto ?

jiradeto: Sure, fixed!

PR #555, opened 2019-06-14
Add support for built-in Matrix functions

jiradeto:
Related issue: #465

Since all matrix built-in functions are already implemented in GraphicsFuzz, this PR mainly focuses on refactoring and documentation.


ast/src/main/java/com/graphicsfuzz/common/glslversion/Glsl330.java
```diff
@@ -34,7 +34,7 @@ public String getVersionString() {
   public boolean supportedDeterminant() {
     // According to this page:
     // https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/determinant.xhtml
-    // determinant should be supported from GLSL 1.40, but glslangValidator disagrees.
```
afd: Can you double-check this?  I think glslangValidator does allow it from version 1.50, so I probably just made an error.  Can you please use glslangValidator and also refer to the shading language versions to check what is the right thing to do here?

ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -172,6 +172,15 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
    */
   boolean supportedDerivativeFunctions();
 
+  /**
```
afd: As above, I think perhaps this comment can be revised or removed.

ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -215,6 +224,15 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
    */
   boolean supportedInterpolationFunctions();
 
+  /**
```
afd: Seems that inverse is supported by glslangValidator in GLSL 1.5.

ast/src/main/java/com/graphicsfuzz/common/glslversion/Glsl330.java
jiradeto: You are right, glslangValidator now does allow this built-in function from version 1.50.

ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -234,9 +246,15 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
   boolean supportedMixFloatBool();
 
   boolean supportedMixNonfloatBool();
-
```
afd: Please avoid adding whitespace like on the following line.

PR #558, opened 2019-06-14
Use try-with-resources to ensure file is closed.

### LINKED ISSUE #556: GraphicsFuzz tests fail on OpenJDK 8 under Windows
The failure is due to an attempt to move a file that seems to still be open.

This seems to be due to the way the .json file for a shader job is read when the shader job is parsed.

afd:
Fixes #556, whereby some tests would fail under Windows with OpenJDK
due to an attempt to move a file that was still open.


paulthomson:
If you search all *.java files for "Reader" or "Writer", there are many MANY cases of this type of thing. Shall we fix them all now? We could add a utility function for reading JSON like this.

Issue #519, opened 2019-06-05
Reducer: replace Global VariablesDeclarationReduction with expression


jiradeto:
Similar idea to #480:

When not preserving semantics, the reducer should try to replace global veriable declarations:
```c
int a = 1;
void main()
{
}
``` 
with 
```c
int a;
void main()
{
 a = 1;
}
```
Note that the new expressions must be the first statement in `main()`.

jiradeto:
Fixed by #576.

PR #561, opened 2019-06-17
Reducer: Add class skeleton and tests

jiradeto:
Related issue: #519

Added class skeleton and tests for the reducer to reduce global initializers with the assignment statements.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunity.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunity.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019 The GraphicsFuzz Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunity.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.graphicsfuzz.common.ast.visitors.VisitationDepth;
+
+public class GlobalVariableDeclToExprReductionOpportunity extends AbstractReductionOpportunity {
+  GlobalVariableDeclToExprReductionOpportunity(VisitationDepth depth) {
+    super(depth);
+  }
+
+  @Override
+  void applyReductionImpl() {
```
afd: I favour "throw new RuntimeException("Not implemented yet.")"

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunity.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  public boolean preconditionHolds() {
```
afd: Similarly, I would throw an exception here to indicate that it's not implemented.

This reduces the risk that something gets executed despite not being ready.

PR #562, opened 2019-06-17
Add new ways to generate opaque zero/one from bitwise operations

### LINKED ISSUE #551: Write new ways of producing opaque zero/one with bitwise operators
Similar to #548, it'd be nice to have some ways of writing opaque zero/one with bitwise operations.

One thing that should be kept in mind is that precision of integers is not invariant, so bitshifting may be precarious.

abelbriggs1:
Fixes #551.

Possibilities added for opaque zero generation:

(opaque zero) >> (0 if signed, otherwise a number between 0 and 1000)
(opaque zero) << (0 if signed, otherwise a number between 0 and 1000)
(opaque zero or one) & (opaque zero)
(opaque zero) & (opaque zero or one)
(opaque zero) | (opaque zero)
(opaque zero) ^ (opaque zero)
(opaque one) ^ (opaque one)

Possibilities added for opaque one generation:

(opaque one) >> 0
(opaque one) << 0
(opaque one) & (opaque one)
(opaque one) | (opaque zero or one)
(opaque zero or one) | (opaque one)
(opaque zero) ^ (opaque one)
(opaque one) ^ (opaque zero)

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -236,7 +236,7 @@ public final class OpaqueExpressionGenerator {
                   new ParenExpr(
                       new BinaryExpr(
                           makeOpaqueOne(type, constContext, depth, fuzzer),
-                          shiftValue.clone(),
+                          shiftValue,
                           BinOp.SHL)),
                   shiftBackValue,
                   BinOp.SHR)));
@@ -252,10 +252,9 @@ public final class OpaqueExpressionGenerator {
    * to be generated depending on the precision and type of the integer.
    *
    * @param type - the type to make a clamped expression from.
-   * @param bound - the bound for the randomly generated number.
-   * @return an expression of a clamped value between 0 and m, inclusive:
-   *     clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(m)), and m is an integer
-   *     between 0 and maxValue, inclusive.
+   * @param bound - the upper bound for the clamped expression.
+   * @return an expression of a clamped value between 0 and bound, inclusive:
+   *     clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(bound)),
    */
   private Expr makeClampedFuzzedExpr(BasicType type, boolean constContext,
                                      final int depth, Fuzzer fuzzer, int bound) {
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -193,7 +193,7 @@ public final class OpaqueExpressionGenerator {
     final Expr shiftValue;
     if (isZero) {
       // We can't shift zero reliably if the number is signed because of sign bit extension, so
-      // we simply shift it zero.
+      // we simply shift it zero if it's signed.
       shiftValue = type.getElementType() == BasicType.INT
           ? new IntConstantExpr("0")
           : new UIntConstantExpr(generator.nextInt(1000) + "u");
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -167,9 +167,10 @@ public final class OpaqueExpressionGenerator {
   }
 
   /**
-   * Function to generate an opaque zero or one by bitwise shifting left or right by an
-   * arbitrary number of bits. Generates an opaque value if the type of zero to be generated is an
-   * integer and the shading language in use supports bitwise operations.
+   * Function to generate an opaque zero or one by bitwise shifting left or right by a value
+   * between 0 and 8 bits, inclusive. For example, we can transform an integer i -> i >> 0 or
+   * i << 0. Generates an opaque value if the type of zero to be generated is an integer and the
+   * shading language in use supports bitwise operations.
    *
    * @param type - the base type of the opaque value being created.
    * @param constContext - true if we're in a constant expression context, false otherwise.
@@ -189,35 +190,54 @@ public final class OpaqueExpressionGenerator {
       return Optional.empty();
     }
     final BinOp operator = generator.nextBoolean() ? BinOp.SHL : BinOp.SHR;
-    // We need to make sure the LHS and RHS of the shift expression are based on the same type
-    // explicitly, so we can't generate opaque values to shift by as they will implicitly be
-    // considered signed integers, and both sides of the shift must be the same type.
-    final Expr shiftValue;
+    // The maximum safe value to shift zero without potentially causing undefined behavior is
+    // between zero and eight bits (as nine bits is the minimum precision of an integer in GLSL).
+    // If we're shifting one, we want to be able to shift backwards without losing information, and
+    // so our maximum shift value is seven bits.
+    final int bound = isZero ? 10 : 9;
+    final int maxValue = generator.nextInt(bound);
+    final Expr shiftValue = new FunctionCallExpr(
+        "clamp",
+        fuzzer.fuzzExpr(type, false, constContext, depth),
+        makeOpaqueZero(type, constContext, depth, fuzzer),
+        applyIdentityFunction(
+            new TypeConstructorExpr(
+                type.toString(),
+                type.getElementType() == BasicType.INT
+                    ? new IntConstantExpr(String.valueOf(maxValue))
+                    : new UIntConstantExpr(maxValue + "u")),
+            type, constContext, depth, fuzzer));
     if (isZero) {
-      // We can't shift zero reliably if the number is signed because of sign bit extension, so
-      // we simply shift it zero if it's signed.
-      shiftValue = type.getElementType() == BasicType.INT
-          ? new IntConstantExpr("0")
-          : new UIntConstantExpr(generator.nextInt(1000) + "u");
+      return Optional.of(
+          new ParenExpr(
+              new BinaryExpr(
+                  makeOpaqueZero(type, constContext, depth, fuzzer),
+                  applyIdentityFunction(shiftValue, type, constContext, depth, fuzzer),
+                  operator)));
     } else {
       assert !isZero;
-      // We can't shift one at all without changing it.
-      shiftValue = type.getElementType() == BasicType.INT
-          ? new IntConstantExpr("0")
-          : new UIntConstantExpr("0u");
+      assert operator == BinOp.SHL || operator == BinOp.SHR;
+      // We're going to shift twice in opposite directions by the same value.
+      final BinOp secondOp = operator == BinOp.SHL ? BinOp.SHR : BinOp.SHL;
+      return Optional.of(
+          new ParenExpr(
+              new BinaryExpr(
+                  new ParenExpr(
+                      new BinaryExpr(
+                          makeOpaqueOne(type, constContext, depth, fuzzer),
+                          applyIdentityFunction(shiftValue.clone(), type, constContext, depth,
+                              fuzzer),
+                          operator)),
+                  applyIdentityFunction(shiftValue.clone(), type, constContext, depth, fuzzer),
+                  secondOp)));
     }
-    return Optional.of(
-        new ParenExpr(
-            new BinaryExpr(
-                makeOpaqueZeroOrOne(isZero, type, constContext, depth, fuzzer),
-                shiftValue, operator)));
   }
 
   /**
    * Function to generate an opaque value by performing a bitwise operation on an opaque zero or
    * an opaque one.
    * Possibilities to create zero include:
-   * Bitwise ANDing an opaque zero or one with an opaque zero: (opaque zero or one) & (opaque zero)
+   * Bitwise ANDing a fuzzed expression with an opaque zero: (fuzzedexpr) & (opaque zero)
    * Bitwise ORing an opaque zero with an opaque zero: (opaque zero) | (opaque zero)
    * Bitwise XORing an opaque zero with an opaque zero or an opaque one with an opaque one:
    *     (opaque zero) ^ (opaque zero) or (opaque one) ^ (opaque one)
@@ -253,14 +273,12 @@ public final class OpaqueExpressionGenerator {
               new ParenExpr(
                   generator.nextBoolean()
                   ? new BinaryExpr(
-                      makeOpaqueZeroOrOne(generator.nextBoolean(), type, constContext, depth,
-                          fuzzer),
+                      fuzzer.fuzzExpr(type, false, constContext, depth),
                       makeOpaqueZero(type, constContext, depth, fuzzer),
                       BinOp.BAND)
                   : new BinaryExpr(
                       makeOpaqueZero(type, constContext, depth, fuzzer),
-                      makeOpaqueZeroOrOne(generator.nextBoolean(), type, constContext, depth,
-                          fuzzer),
+                      fuzzer.fuzzExpr(type, false, constContext, depth),
                       BinOp.BAND)));
         } else {
           assert !isZero;
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -167,10 +167,24 @@ public final class OpaqueExpressionGenerator {
   }
 
   /**
-   * Function to generate an opaque zero or one by bitwise shifting left or right by a value
-   * between 0 and 8 bits, inclusive. For example, we can transform an integer i -> i >> 0 or
-   * i << 0. Generates an opaque value if the type of zero to be generated is an integer and the
-   * shading language in use supports bitwise operations.
+   * Function to generate an opaque zero or one by bitwise shifting left or right by an amount of
+   * bits dependent on which opaque is being generated.
+   * The minimum precision for a lowp integer in GLSL is 9 bits, with one of those bits reserved as
+   * a sign bit if the integer is signed. The OpenGL specification does not define behavior for
+   * shifting an integer beyond its maximum size in bits.
+   * This influences the process of generating opaque values such that:
+   *     If we're generating an opaque zero, we don't have to worry about losing bits, so our
+   *     maximum shift value is 8 bits.
+   *     If we're generating an opaque one, we have to make sure that we don't lose our 1 bit.
+   *     This limits our maximum shift value to 7 bits (more on how we shift an opaque one below).
+   * Possibilities for generating an opaque zero include:
+   *     Shifting an opaque zero by n bits: (opaque zero) >> n or (opaque zero) << n, where n is
+   *     an expression of a clamped value between 0 and 8:
+   *     n = clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(8)).
+   * Possibilities for generating an opaque zero include:
+   *     Shifting an opaque one to the left by n bits, then shifting it to the right by n bits:
+   *     ((opaque one) << n) >> n, where n is an expression of a clamped value between 0 and 7,
+   *     inclusive: n = clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(7)).
    *
    * @param type - the base type of the opaque value being created.
    * @param constContext - true if we're in a constant expression context, false otherwise.
@@ -189,16 +203,21 @@ public final class OpaqueExpressionGenerator {
     if (!shadingLanguageVersion.supportedBitwiseOperations()) {
       return Optional.empty();
     }
-    final BinOp operator = generator.nextBoolean() ? BinOp.SHL : BinOp.SHR;
-    // The maximum safe value to shift zero without potentially causing undefined behavior is
-    // between zero and eight bits (as nine bits is the minimum precision of an integer in GLSL).
-    // If we're shifting one, we want to be able to shift backwards without losing information, and
-    // so our maximum shift value is seven bits.
-    final int bound = isZero ? 10 : 9;
-    final int maxValue = generator.nextInt(bound);
+    // The minimum precision for a lowp integer in GLSL is 9 bits (with one reserved for a sign
+    // bit if the integer is signed) - we don't have to worry about losing information if we're
+    // shifting zero, but shifting more than 8 bits may result in undefined behavior.
+    final int maxShiftForZero = 9;
+    // While we still have a hard maximum bits to shift of 8 bits, if we're shifting one, then
+    // we can potentially lose information because GLSL bit shifting is not circular. To remedy
+    // that, we make sure not to shift our 1 bit out of the integer by limiting our maximum shift
+    // to 7 bits.
+    final int maxShiftForOne = 8;
+    final int maxValue = generator.nextInt(isZero ? maxShiftForZero : maxShiftForOne);
+    // We pass true as constContext when fuzzing here because the expression will be evaluated,
+    // so we don't want any side effects.
     final Expr shiftValue = new FunctionCallExpr(
         "clamp",
-        fuzzer.fuzzExpr(type, false, constContext, depth),
+        fuzzer.fuzzExpr(type, false, true, depth),
         makeOpaqueZero(type, constContext, depth, fuzzer),
         applyIdentityFunction(
             new TypeConstructorExpr(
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
         depth, fuzzer)));
   }
 
+  /**
+   * Function to generate an opaque zero or one by bitwise shifting left or right by an
+   * arbitrary number of bits. Generates an opaque value if the type of zero to be generated is an
+   * integer and the shading language in use supports bitwise operations.
```
afd: Can you add a couple of intuitive examples to the comment?

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
         depth, fuzzer)));
   }
 
+  /**
+   * Function to generate an opaque zero or one by bitwise shifting left or right by an
```
afd: I'm not sure it's right to say that an arbitrary number of bits is used here.  In the code you take some care about the shift value (and I think maybe even more care needs to be taken; see comments below).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+                                                     final int depth, Fuzzer fuzzer,
+                                                     boolean isZero) {
+    if (!BasicType.allIntegerTypes().contains(type)) {
+      return Optional.empty();
+    }
+    if (!shadingLanguageVersion.supportedBitwiseOperations()) {
+      return Optional.empty();
+    }
+    final BinOp operator = generator.nextBoolean() ? BinOp.SHL : BinOp.SHR;
+    // We need to make sure the LHS and RHS of the shift expression are based on the same type
```
afd: The GLSL ES 3.20 spec says:

"One operand can be signed while the other is unsigned. In all cases, the resulting type will be the same type as the left operand."

So I suggest that you randomly choose whether the RHS is signed or unsigned.

afd: Also, the spec says:

"If the first operand is a vector, the second operand
must be a scalar or a vector with the same size as the first operand, and the result is computed
component-wise."

So I suggest that you randomly choose whether the second argument should be scalar or vector, such that if it is chosen to be vector it has the same number of elements as the LHS (so that if the LHS is already a scalar, the RHS is a scalar either way).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    // explicitly, so we can't generate opaque values to shift by as they will implicitly be
+    // considered signed integers, and both sides of the shift must be the same type.
+    final Expr shiftValue;
+    if (isZero) {
```
afd: I think we're OK here actually.  The spec says:

"If E1 is a signed integer, the right-shift will extend the sign bit. If E1 is an
unsigned integer, the right-shift will zero-extend."

Given that the sign bit of 0 is 0, right-shifting will maintain zero-ness even in the signed case.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      // We can't shift zero reliably if the number is signed because of sign bit extension, so
+      // we simply shift it zero if it's signed.
+      shiftValue = type.getElementType() == BasicType.INT
+          ? new IntConstantExpr("0")
```
afd: I don't think it is safe to use a random value in the range [0, 1000).  The spec says:

"The result is undefined if the right operand is negative, or greater than or
equal to the number of bits in the left expression’s base type."

Now, what can we assume about the number of bits in the left expression's base type?  Without some precision analysis, we should conservatively assume that it might be a lowp integer.  The spec says:

"lowp integers (signed and unsigned) must be represented as an integer with between 9 and 32 bits
inclusive."

So to be safe we should make the RHS be in the range [0, 9).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+          : new UIntConstantExpr(generator.nextInt(1000) + "u");
```
afd: Now, rather than generating a literal in the range [0, 9], it would be more adventurous to generate a random expression and clamp it into the range [0, 9).

I would do something like:

identity(clamp(fuzzed_expr, opaque_zero, identity(9))

for the RHS.

afd: A problem with clamping into the range [0, 9) is that the result will almost certainly be 8.  An alternative that might be better is to pick a value n, with 0 <= n <= 9, and clamp into the range [0, n).  The result will almost certainly be (n-1), but since n is chosen randomly this creates more variation.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    } else {
+      assert !isZero;
+      // We can't shift one at all without changing it.
```
afd: What we could do is left-shift it by n bits and then right-shift it by n bits, where n <= 7.  My thinking is that:

- we know the value will be at least 9 bits wide
- 1 << 7 is has sign bit 0 when interpreted as a 9-bit integer
- Thus (1 << 7) >> 7 will still be 1.

If you do this then I suggest choosing the shift distance randomly, and then generating a fuzzed expression and clamping it into the right range.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,157 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+            new BinaryExpr(
+                makeOpaqueZeroOrOne(isZero, type, constContext, depth, fuzzer),
+                shiftValue, operator)));
+  }
+
+  /**
+   * Function to generate an opaque value by performing a bitwise operation on an opaque zero or
+   * an opaque one.
+   * Possibilities to create zero include:
+   * Bitwise ANDing an opaque zero or one with an opaque zero: (opaque zero or one) & (opaque zero)
```
afd: ^ Why not bitwise AND a fuzzed expression with an opaque zero?  (More general.)  And I presume you're doing it in both directions.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Got it.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: You're correct - it seems I was under the wrong impression from an error I got while testing that was unrelated to this.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: You're correct, I think the issue I had when testing was unrelated to this, but I thought that this was the cause.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Now that we're clamping, changing the type in the clamp starts to cause issues with applying other identity functions to the resulting expressions (specifically, ternary type mismatches).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    }
+    final BinOp operator = generator.nextBoolean() ? BinOp.SHL : BinOp.SHR;
+    // The maximum safe value to shift zero without potentially causing undefined behavior is
+    // between zero and eight bits (as nine bits is the minimum precision of an integer in GLSL).
+    // If we're shifting one, we want to be able to shift backwards without losing information, and
+    // so our maximum shift value is seven bits.
+    final int bound = isZero ? 10 : 9;
+    final int maxValue = generator.nextInt(bound);
+    final Expr shiftValue = new FunctionCallExpr(
+        "clamp",
```
afd: I think we should pass 'true' as the value of 'constContext' in the call to 'fuzzExpr', as we don't want to end up fuzzing an expression that may have side-effects.  This is important because the expression is really going to be evaluated here.

(When thinking about this, I ended up opening #570 because we would like to generate fuzzed expressions that use builtins to some degree.)

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+                    : new UIntConstantExpr(maxValue + "u")),
+            type, constContext, depth, fuzzer));
+    if (isZero) {
+      return Optional.of(
+          new ParenExpr(
+              new BinaryExpr(
+                  makeOpaqueZero(type, constContext, depth, fuzzer),
+                  applyIdentityFunction(shiftValue, type, constContext, depth, fuzzer),
+                  operator)));
+    } else {
```
afd: I would remove the following assertion as it is the negation of the enclosing 'if' condition.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      assert !isZero;
```
afd: I also wouldn't bother with the following assertion, since you didn't assert it in the 'then' branch and there seems no special reason to assert it here (though it does not harm I suppose).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      assert operator == BinOp.SHL || operator == BinOp.SHR;
+      // We're going to shift twice in opposite directions by the same value.
```
afd: The only legal choice is to shift left then shift right.  The problem is that if we shift right first, the 1 will get shifted away and we'll end up with 0.  (I don't think it is worth allowing this for the special case where the shift value is 0.)

I suggest you choose the shift operator in the "isZero" branch, and then change this branch so that the order of operators is fixed.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+                      new BinaryExpr(
+                          makeOpaqueOne(type, constContext, depth, fuzzer),
+                          applyIdentityFunction(shiftValue.clone(), type, constContext, depth,
+                              fuzzer),
+                          operator)),
+                  applyIdentityFunction(shiftValue.clone(), type, constContext, depth, fuzzer),
+                  secondOp)));
+    }
+  }
+
```
afd: ^ A general thing on the above: I'm a little worried about getting expression size explosion due to the amount of places the identity function is applied.  Perhaps don't apply the identity function to "shiftValue.clone()".

A separate point: it seems nice to have the option to shift left and then right using the same expression as shift value.  But it would also be interesting to shift left with one expression that is clamped to n, and right with a different expression that is clamped to n.  How about you decide with 0.5 probability whether to generate a new shift value expression for shifting back, vs. cloning the existing one?

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+                type.getElementType() == BasicType.INT
+                    ? new IntConstantExpr(String.valueOf(maxValue))
+                    : new UIntConstantExpr(maxValue + "u")),
+            type, constContext, depth, fuzzer));
+    if (isZero) {
+      return Optional.of(
+          new ParenExpr(
+              new BinaryExpr(
+                  makeOpaqueZero(type, constContext, depth, fuzzer),
+                  applyIdentityFunction(shiftValue, type, constContext, depth, fuzzer),
```
afd: ^ On second thoughts, let's not apply identity to shiftValue; it's already a fuzzed expression and we don't want to go too crazy with identity applications.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+   *     made from performing a bitwise operation on an opaque zero or opaque one.
+   */
+  private Optional<Expr> opaqueZeroOrOneBitwiseOp(BasicType type, boolean constContext,
+                                                   final int depth, Fuzzer fuzzer, boolean isZero) {
+    if (!BasicType.allIntegerTypes().contains(type)) {
+      return Optional.empty();
+    }
+    if (!shadingLanguageVersion.supportedBitwiseOperations()) {
+      return Optional.empty();
+    }
```
afd: Can you replace '3' with a final int with a suitable name, capturing the fact that this is the number of options?

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    final int operator = generator.nextInt(3);
+    Optional<Expr> opaqueExpr;
+    switch (operator) {
+      case 0:
+        // Bitwise AND
+        if (isZero) {
+          opaqueExpr = Optional.of(
+              new ParenExpr(
+                  generator.nextBoolean()
+                  ? new BinaryExpr(
```
afd: When passing 'constContext' to 'fuzzExpr', you should pass the literal 'true', because these fuzzed expressions really wil get evaluated (no short-circuit evaluation for bitwise operators).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+              new ParenExpr(
+                  new BinaryExpr(
+                      makeOpaqueZero(type, constContext, depth, fuzzer),
+                      makeOpaqueZero(type, constContext, depth, fuzzer),
+                      BinOp.BOR)));
+        } else {
+          assert !isZero;
+          opaqueExpr = Optional.of(
+              new ParenExpr(
+                  generator.nextBoolean()
```
afd: It would be more concise here to introduce two variables, one to hold the result of makeOpaqueOne and the other the result of makeOpaqueZeroOrOne, to factor out the common expressions.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -164,6 +166,175 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+                      BinOp.BOR)
+                  : new BinaryExpr(
+                      makeOpaqueZeroOrOne(generator.nextBoolean(), type, constContext, depth,
+                          fuzzer),
+                      makeOpaqueOne(type, constContext, depth, fuzzer),
+                      BinOp.BOR)));
+        }
+        break;
+      default:
+        // Bitwise XOR
```
afd: Change the '2' to (v - 1) where v is the variable you introduced to express the maximum number of options.

abelbriggs1:
I fixed up the comments and generally refactored according to your feedback.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -167,10 +167,24 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+   *     If we're generating an opaque one, we have to make sure that we don't lose our 1 bit.
+   *     This limits our maximum shift value to 7 bits (more on how we shift an opaque one below).
+   * Possibilities for generating an opaque zero include:
+   *     Shifting an opaque zero by n bits: (opaque zero) >> n or (opaque zero) << n, where n is
+   *     an expression of a clamped value between 0 and 8:
+   *     n = clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(8)).
+   * Possibilities for generating an opaque zero include:
+   *     Shifting an opaque one to the left by n bits, then shifting it to the right by n bits:
+   *     ((opaque one) << n) >> n, where n is an expression of a clamped value between 0 and 7,
+   *     inclusive: n = clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(7)).
```
afd: The above is really excellent!

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -189,16 +203,21 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
-    final BinOp operator = generator.nextBoolean() ? BinOp.SHL : BinOp.SHR;
-    // The maximum safe value to shift zero without potentially causing undefined behavior is
-    // between zero and eight bits (as nine bits is the minimum precision of an integer in GLSL).
-    // If we're shifting one, we want to be able to shift backwards without losing information, and
-    // so our maximum shift value is seven bits.
-    final int bound = isZero ? 10 : 9;
-    final int maxValue = generator.nextInt(bound);
+    // The minimum precision for a lowp integer in GLSL is 9 bits (with one reserved for a sign
+    // bit if the integer is signed) - we don't have to worry about losing information if we're
+    // shifting zero, but shifting more than 8 bits may result in undefined behavior.
```
afd: The name "maxShiftForZero" isn't right; the max shift for zero is actually 8.  I would call the variable "minBitsForLowpInt".

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -189,16 +203,21 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    final int maxShiftForZero = 9;
+    // While we still have a hard maximum bits to shift of 8 bits, if we're shifting one, then
+    // we can potentially lose information because GLSL bit shifting is not circular. To remedy
+    // that, we make sure not to shift our 1 bit out of the integer by limiting our maximum shift
+    // to 7 bits.
```
afd: Then I would rename this one to "minNonSignBitsForLowpInt", and initialize it to "minBitsForLowpInt - 1" (rather than explicitly to 7).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -208,40 +227,37 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
           new ParenExpr(
               new BinaryExpr(
                   makeOpaqueZero(type, constContext, depth, fuzzer),
-                  applyIdentityFunction(shiftValue, type, constContext, depth, fuzzer),
+                  shiftValue,
                   operator)));
     } else {
-      assert !isZero;
-      assert operator == BinOp.SHL || operator == BinOp.SHR;
       // We're going to shift twice in opposite directions by the same value.
```
afd: How about here you do:

final Expr shiftBackValue = generaor.nextBool() ? shiftValue.clone() : (another fuzzed expression clamped to maxValue);

Then below change the first shiftValue.clone() to shiftValue, and the second shiftValue.clone() to shiftBackValue.

This will be nice as it will sometimes lead to some duplicated expressions (giving the compiler room for common subexpression elimination optimizations), and other times leads to very different but equivalent expressions (testing the compiler in other ways).

As you'll be generating the big clamped expression twice, it would be worth factoring that out into a private utility method.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Thanks, it took awhile for me to figure out how I wanted to phrase it.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Done

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -246,11 +238,41 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
                           makeOpaqueOne(type, constContext, depth, fuzzer),
```
afd: There's no need to clone shiftValue here, as it's not used multiple times.  (You needed to clone it before as you were using it twice).

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -246,11 +238,41 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+   * Utility function to clamp a fuzzed expression between an opaque zero and the identity of a
+   * type constructor of the given type, with the value of the bound argument. Note that this
+   * function only supports integer types currently - it could be extended to support floating
+   * point numbers as well if needed. Another note is that this function does not check its bound
+   * for validity - specifying a bound larger than 256 or a negative value could cause invalid GLSL
+   * to be generated depending on the precision and type of the integer.
+   *
+   * @param type - the type to make a clamped expression from.
+   * @param bound - the bound for the randomly generated number.
+   * @return an expression of a clamped value between 0 and m, inclusive:
```
afd: I guess you mean "bound" rather than "maxValue" below?

abelbriggs1: Oops!

PR #565, opened 2019-06-18
Clean up for Amber limitation of a single type per SSBO

hevrard:


python/src/main/python/drivers/runspv.py
```diff
@@ -1109,12 +1103,21 @@ def comp_json_to_amberscript(comp_json):
 
     binding = j['buffer']['binding']
     offset = 0
+
+    # Amber only support one type per buffer, check this limitation.
```
afd: ^ support -> supports

PR #576, opened 2019-06-20
Reducer implementation: reduce global variable declarations to expression 

jiradeto:
Related issue: #519

Implemented new opportunities to remove initializers from global variable declarations and replace them with assignment statements.


jiradeto:
Well, if we do not retain the order of variable assignment statements, would the semantics of the code get changed?

For example: 
```c
int a = 1;
int b = a;
void main(){
}
```
to 
```c
int a;
int b;
void main(){
  b = a; // assignment variable b is inserted before a
  a = 1;
}
```

paulthomson:
I can't speak for Ally but I think I can make two points:

1. The reducer will pick *randomly* from the opportunities list. Thus, the reducer will *not* respect the order of the list.
2. I believe this reduction opportunity is assumed to be semantics-*changing*, according to #519.

afd:
That's correct, Paul.

@jiradeto as Paul says, even if you do create the opportunities in a certain order, the reducer might try only a few of them; or it might try all of them but in a different order.  So there's no way we can guarantee preservation of semantics, and that's why we don't collect any opportunities unless the reducer is running in semantics-changing mode.

Hope that makes sense!

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunities.java
```diff
@@ -17,18 +17,55 @@
...
+ * int b;
+ * void main(){
+ *  a = 1;
+ *  b = foo();
+ *  int c = 1;
+ * }
+ */
+
 public class GlobalVariableDeclToExprReductionOpportunities
     extends ReductionOpportunitiesBase<GlobalVariableDeclToExprReductionOpportunity> {
```
afd: I favour initializing fields in the constructor rather than at the declaration.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunities.java
```diff
@@ -46,4 +83,41 @@ public GlobalVariableDeclToExprReductionOpportunities(TranslationUnit tu,
...
+      globalVariableDecl.add(variablesDeclaration);
+    }
+  }
+
+  @Override
+  public void visitFunctionDefinition(FunctionDefinition functionDefinition) {
+    super.visitFunctionDefinition(functionDefinition);
+    // We consider only the global variable declarations with the initializer that have been
+    // found before main function.
+    if (functionDefinition.getPrototype().getName().equals("main")) {
```
afd: There's no need to have the explicit indices - as discussed in the comments earlier, there's no need to - and indeed no point - enforcing an ordering in which the opportunities are added.  So you should change this to:

for (VariablesDeclaration variablesDeclaration : globalVariableDecl) {
   for (VariableDeclInfo variableDeclInfo : variablesDeclaration.getDeclInfos()) {
      if (...) {
        ...
      }
   }
}

It's best to avoid stuff with explicit indices where possible.

jiradeto: Since in the [previous PR](https://github.com/google/graphicsfuzz/pull/561/files#diff-901b1045d425182b8e372343e60e8227R179) there is a test checking the order of the variable assignment, if we do not enforce the order of the assignment here test will fail. So should I remove such test in this PR?

afd: Ah, I suggest you change that test so that it works for the ordering that the above reformulation will yield.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/GlobalVariableDeclToExprReductionOpportunities.java
```diff
@@ -17,18 +17,56 @@
...
+ * }
+ */
+
 public class GlobalVariableDeclToExprReductionOpportunities
     extends ReductionOpportunitiesBase<GlobalVariableDeclToExprReductionOpportunity> {
+  private final List<VariablesDeclaration> globalVariableDecl;
+
   public GlobalVariableDeclToExprReductionOpportunities(TranslationUnit tu,
                                                         ReducerContext context) {
     super(tu, context);
```
afd: I favour always using "this." in a constructor, just for uniformity.

jiradeto: Fixed

PR #577, opened 2019-06-24
Add script to convert a GraphicsFuzz shader job to a piglit shader test

### LINKED ISSUE #571: Develop script to convert a shader job into a piglit/shader_runner test
When reporting bugs for Mesa, it's helpful for the developers to have an easy-to-run test. GraphicsFuzz takes some effort to set up - instead, [piglit](https://gitlab.freedesktop.org/mesa/piglit) is often used to report errors.

It takes some effort to figure out how exactly to create a piglit test - it'd be helpful to have a script that automates creating a piglit .shader_test file from a GraphicsFuzz shader job.

abelbriggs1:
Fixes #571.

Given a shader job JSON file as an argument (with a corresponding shader file of the same name in the same directory), this script outputs a piglit shader_runner test to the same directory that loads the uniforms in the JSON file, runs the test, and draws the image.

As I don't know how compute shader data is loaded into piglit (and I couldn't find docs on it), this script currently only works for fragment shaders.

My Python is pretty rusty, so feel free to be picky about code review.

### DIFF
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -187,7 +187,7 @@ def get_shader_from_job(shader_job: str) -> str:
     :param shader_job: the path of the shader job file.
     :return: the path of the fragment shader file.
     """
-    return remove_end(shader_job, '.json') + '.frag'
+    return runspv.remove_end(shader_job, '.json') + '.frag'
 
 
 def get_shader_test_from_job(shader_job: str) -> str:
@@ -196,7 +196,7 @@ def get_shader_test_from_job(shader_job: str) -> str:
     :param shader_job: the path of the shader job file.
     :return: the path of the shader_test file.
     """
-    return remove_end(shader_job, '.json') + '.shader_test'
+    return runspv.remove_end(shader_job, '.json') + '.shader_test'
 
 
 def main_helper(args: List[str]) -> None:
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -46,6 +46,18 @@ ES_SPECIFIER = ' es'
 # GLSL version headers specify a version without a decimal point, when piglit takes a version
 # string with a decimal point. The easiest way of getting this is by dividing by 100.
 SHADER_VERSION_FACTOR = 100
+# Uniform types to be found in the JSON.
+UNIFORM_TYPES = {
+    'glUniform1i': 'int',
+    'glUniform2i': 'ivec2',
+    'glUniform3i': 'ivec3',
+    'glUniform4i': 'ivec4',
+    'glUniform1f': 'float',
+    'glUniform2f': 'vec2',
+    'glUniform3f': 'vec3',
+    'glUniform4f': 'vec4',
+}
+UNIFORM_DEC = 'uniform'
 
 
 def make_shader_test_string(shader_job: str) -> str:
@@ -68,7 +80,7 @@ def make_shader_test_string(shader_job: str) -> str:
     shader_test_string += make_require_header(shader_version_header) + '\n'
     shader_test_string += make_vertex_shader_header() + '\n'
     shader_test_string += make_fragment_shader_header(shader_file_string) + '\n'
-    shader_test_string += make_test_header(shader_job_json_parsed, shader_lines)
+    shader_test_string += make_test_header(shader_job_json_parsed)
 
     return shader_test_string
 
@@ -117,36 +129,25 @@ def make_fragment_shader_header(fragment_shader: str) -> str:
     return frag_header
 
 
-def make_test_header(shader_job_json_parsed: List, shader_lines: List) -> str:
+def make_test_header(shader_job_json_parsed: dict) -> str:
     """
     Creates the [test] header. Loads uniforms based on the uniforms found in the shader.
     Note that piglit really doesn't like it if you try to load a uniform that isn't actually
     used in the shader - because of this, we actually have to parse the shader itself and
     find out which uniforms are being used instead of blindly pasting from the shader job JSON file.
     :param shader_job_json_parsed: the parsed JSON properties.
-    :param shader_lines: the shader code parsed into a list.
     :return: the shader_test test header string.
     """
     test_header = TEST_HEADER + '\n'
-    uniforms_in_shader = get_uniforms_from_shader(shader_lines)
-    for dec in uniforms_in_shader:
-        dec = dec.strip(';')
-        test_header += dec + ' '
-        variable_name = dec.split(' ')[-1]
-        if variable_name not in shader_job_json_parsed:
-            raise IOError('Malformed shader job - ' + dec + ' not found in job.')
-        uniform_args = shader_job_json_parsed.get(variable_name).get('args')
-        test_header += str(uniform_args[0:]).strip('[]').replace(',', '') + '\n'
+    for key, value in shader_job_json_parsed.items():
+        test_header += UNIFORM_DEC + ' '
+        uniform_type = value['func']
+        if uniform_type not in UNIFORM_TYPES.keys():
+            raise AssertionError('Unknown uniform type: ' + uniform_type)
+        test_header += UNIFORM_TYPES[uniform_type] + ' ' + str(key) + ' '
+        test_header += str(value['args'][0:]).strip('[]').replace(',', '') + '\n'
     test_header += DRAW_COMMAND
     return test_header
-
-
-def get_uniforms_from_shader(shader_lines: List) -> List:
-    """
-    :param shader_lines: the shader code parsed into a list
-    :return: a list of all uniform declarations in the shader
-    """
-    return (line for line in shader_lines if is_uniform(line))
 
 
 def is_uniform(line: str) -> bool:
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -14,9 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import runspv
 import argparse
 import json
-import os
 from typing import List
 
 # Note: We define a 'shader job' as the JSON file of a GraphicsFuzz shader.
@@ -197,17 +197,6 @@ def get_shader_test_from_job(shader_job: str) -> str:
     :return: the path of the shader_test file.
     """
     return remove_end(shader_job, '.json') + '.shader_test'
-
-
-def remove_end(str_in: str, str_end: str) -> str:
-    """
-    Helper function to remove the file suffix of a path. Imported from runspv.py.
-    :param str_in: the string to remove the end from.
-    :param str_end: the end of the string.
-    :return: str_in with the end of the string removed.
-    """
-    assert str_in.endswith(str_end), 'Expected {} to end with {}'.format(str_in, str_end)
-    return str_in[:-len(str_end)]
 
 
 def main_helper(args: List[str]) -> None:
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -17,7 +17,6 @@
 import argparse
 import json
 import os
-import sys
 from typing import List
 
 # Note: We define a 'shader job' as the JSON file of a GraphicsFuzz shader.
@@ -37,8 +36,6 @@ DRAW_COMMAND = 'draw rect -1 -1 2 2'
 
 # Strings used to specify the GL version to use in a piglit test's
 # [require] header.
-GL_VERSION_STRING = 'GL >= '
-GLES_VERSION_STRING = 'GL ES >= '
 GLSL_VERSION_STRING = 'GLSL >= '
 GLSLES_VERSION_STRING = 'GLSL ES >= '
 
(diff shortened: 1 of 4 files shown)


python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+    argparser = argparse.ArgumentParser(description = description)
+
+    argparser.add_argument(
+        'shader_job', 
+        help = ('Path to the GraphicsFuzz shader job JSON file.'))
+    
+    args = argparser.parse_args(args)
+
+    # Make sure the shader job exists and that the given path to write to is valid.
+    check_shader_job_exists(args.shader_job)
```
paulthomson: Nit: it is common in our code base to do a *basic* check of all arguments, so that simple errors are caught early. So this is good in that it follows our style. However, in this case, I would suggest removing this, as (a) the check is somewhat non-trivial and (b) the first thing `make_shader_test_string` does is read the JSON and then read the frag file; if these fail, exceptions will be raised. The Python approach is to just do things and see if an exception is raised, instead of `if check: do thing`. We don't always follow this though, and this is a subjective!

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+  
+    test_string = make_shader_test_string(args.shader_job)
+
+    shader_test_file = open(get_shader_test_from_job(args.shader_job), 'w',
```
paulthomson: I would again use `with` and `errors = "ignore"` here

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+    
+    args = argparser.parse_args(args)
+
+    # Make sure the shader job exists and that the given path to write to is valid.
+    check_shader_job_exists(args.shader_job)
+  
+    test_string = make_shader_test_string(args.shader_job)
+
+    shader_test_file = open(get_shader_test_from_job(args.shader_job), 'w', 
+        encoding = 'utf-8')
```
paulthomson: Finally, I would stick with either `"` or `'` for all strings in this file. If you want to try a code formatter (which will also fix all strings to use `"`), I suggest the `black` Python code formatter. We don't use it yet in GraphicsFuzz, but we will soon. Its main selling point is it cannot be configured :) Although it actually can a bit!

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+def get_shader_from_job(shader_job: str) -> str:
+    return shader_job.split(".")[0] + ".frag"
+
+# Helper function to get the filename of the new shader test from the given shader job.
+def get_shader_test_from_job(shader_job: str) -> str:
+    return shader_job.split(".")[0] + ".shader_test"
+
+def main_helper(args: List[str]) -> int:
+    description = (
+        'Given a GraphicsFuzz shader job JSON file, produce a Mesa piglit \
```
paulthomson: Nit: I believe you can just close the string `'` and re-open it on the next line `'` instead of using `\`, as long as you have parentheses.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+VERTEX_HEADER = '[vertex shader passthrough]'
+FRAGMENT_HEADER = '[fragment shader]'
+TEST_HEADER = '[test]'
+
+# Draw command for piglit to draw the shader's output. Required in the test header.
+DRAW_COMMAND = 'draw rect -1 -1 2 2'
+
+# Strings used to specify the GL version to use in a piglit test's
+# [require] header.
+GL_VERSION_STRING = 'GL >= '
```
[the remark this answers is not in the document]
paulthomson: Hmm actually, it does work most of the time. If you want to keep this, perhaps add a check for if the GLSL version is >= 330, or if it is GLSL ES then check if the version is >= 300 es. Then you can do the dividing trick. Otherwise, I would just not output any GL nor GLES requirements. I assume that these requirements help piglit decide whether the tests make sense to run on the current driver, but I am not sure. I assume `GLSL ES >= x` implies that GLES is required? Not sure though.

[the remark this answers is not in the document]
paulthomson: Looking at shader_runner, I think it checks which versions of GLSL are supported by the driver, so just the GLSL (ES) requirements should be sufficient. No need for the GL or GLES requirements.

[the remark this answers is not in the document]
abelbriggs1: Thanks, I've removed it.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+
+# GLSL preprocessor version flag.
+SHADER_VERSION_FLAG = '#version'
+# GLES specifier in version string.
+ES_SPECIFIER = ' es'
+# GLSL version headers specify a version without a decimal point, when piglit takes a version
+# string with a decimal point. The easiest way of getting this is by dividing by 100.
+SHADER_VERSION_FACTOR = 100
+
+# Makes a piglit shader test from a shader job and shader.
```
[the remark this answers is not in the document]
abelbriggs1: Thanks, I was working on VSCode before - I didn't know IntelliJ had a Python plugin. I've since switched to that.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+# the same directory.
+
+# Piglit test section headers
+REQUIRE_HEADER = '[require]'
+VERTEX_HEADER = '[vertex shader passthrough]'
+FRAGMENT_HEADER = '[fragment shader]'
+TEST_HEADER = '[test]'
+
+# Draw command for piglit to draw the shader's output. Required in the test header.
+DRAW_COMMAND = 'draw rect -1 -1 2 2'
```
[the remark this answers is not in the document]
abelbriggs1: I'll be honest, I have no idea how the draw command really works - I pretty much copy pasted the command from the piglit developer's tutorial blogpost (the only user documentation that exists for the framework) and it seems to draw images properly in the tests I've made so far.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: Fair enough - when I was originally writing this I planned to let the user choose their own directory to write to as a second argument, but when I dropped that I still left this check in.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: I've fixed everything to use ' for strings and """ blocks for function comments.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: Thanks, I wasn't aware of that.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+
+    shader_file = open(get_shader_from_job(shader_job), 'r', 
+        encoding = 'utf-8')
+    shader_file_string = shader_file.read()
+    shader_line_list = shader_file_string.split('\n')
+    shader_file.close()
+    
+    shader_test_string = str('')
+    try:
+        shader_version_header = get_version_header_from_shader(shader_line_list)
```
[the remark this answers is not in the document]
abelbriggs1: I wasn't quite sure, so I made sure to check for it - thanks for clarifying.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,191 @@
...
+def make_fragment_shader_header(fragment_shader: str) -> str:
+    frag_header = FRAGMENT_HEADER + '\n'
+    frag_header += fragment_shader
+    return frag_header
+
+# Creates the [test] header. Loads uniforms based on the uniforms found in the shader.
+# Note that piglit really doesn't like it if you try to load a uniform that isn't actually
+# used in the shader - because of this, we actually have to parse the shader itself and
+# find out which uniforms are being used instead of blindly pasting from the shader job JSON file.
+def make_test_header(shader_job_json_parsed: List, shader_line_list: List) -> str:
```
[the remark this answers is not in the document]
abelbriggs1: One thing to note, though, is that I can't infer the type of the uniform from the JSON - I still have to parse the shader to copypaste the declaration. A solution that manually formats a declaration without parsing the shader might be more complex than with parsing.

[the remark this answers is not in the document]
abelbriggs1: If I didn't parse the shader, though, I'd have to pick apart the 'gluniform2' string and manually format the declaration (uniform vec2 resolution, for instance), which will probably be more complex than this solution (which basically just copy pastes values from the shader and JSON without checking)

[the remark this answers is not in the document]
paulthomson: Yes, that is indeed how we do it when converting to amber script. Perhaps you can re-use code from runspv.py that already does this.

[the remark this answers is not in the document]
paulthomson: Oh actually it is a bit different, but you can see the dict that maps between gl calls and types.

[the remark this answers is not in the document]
abelbriggs1: I've patched it up to only use the JSON - thanks for the reference to runspv.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+    frag_header += fragment_shader
+    return frag_header
+
+
+def make_test_header(shader_job_json_parsed: dict) -> str:
+    """
+    Creates the [test] header. Loads uniforms based on the uniforms found in the shader.
+    Note that piglit really doesn't like it if you try to load a uniform that isn't actually
+    used in the shader - because of this, we actually have to parse the shader itself and
+    find out which uniforms are being used instead of blindly pasting from the shader job JSON file.
```
paulthomson: Update comment :)

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+        if uniform_type not in UNIFORM_TYPES.keys():
+            raise AssertionError('Unknown uniform type: ' + uniform_type)
+        test_header += UNIFORM_TYPES[uniform_type] + ' ' + str(key) + ' '
+        test_header += str(value['args'][0:]).strip('[]').replace(',', '') + '\n'
+    test_header += DRAW_COMMAND
+    return test_header
+
+
+def is_uniform(line: str) -> bool:
+    """
```
paulthomson: You can remove this now.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+    :return: the shader_test test header string.
+    """
+    test_header = TEST_HEADER + '\n'
+    for key, value in shader_job_json_parsed.items():
+        test_header += UNIFORM_DEC + ' '
+        uniform_type = value['func']
+        if uniform_type not in UNIFORM_TYPES.keys():
+            raise AssertionError('Unknown uniform type: ' + uniform_type)
+        test_header += UNIFORM_TYPES[uniform_type] + ' ' + str(key) + ' '
+        test_header += str(value['args'][0:]).strip('[]').replace(',', '') + '\n'
```
paulthomson: I would suggest something like this instead: 

```python
' '.join([str(arg) for arg in value['args']]) + '\n'
```

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+    Creates the [test] header. Loads uniforms based on the uniforms found in the shader.
+    Note that piglit really doesn't like it if you try to load a uniform that isn't actually
+    used in the shader - because of this, we actually have to parse the shader itself and
+    find out which uniforms are being used instead of blindly pasting from the shader job JSON file.
+    :param shader_job_json_parsed: the parsed JSON properties.
+    :return: the shader_test test header string.
+    """
+    test_header = TEST_HEADER + '\n'
+    for key, value in shader_job_json_parsed.items():
+        test_header += UNIFORM_DEC + ' '
```
paulthomson: This is somewhat subjective, but I prefer to use string templates for this kind of thing. This way, it is easy to see the overall structure of what is being output as opposed to many string concatenations.

```python
test_header += 'uniform {type} {uniform_name} {args}\n'.format(
    type=get_uniform_type_from_gl_func(value['func']),
    uniform_name=uniform_name,
    args=' '.join([str(arg) for arg in value['args']]),
)
```

I suggest renaming `key` to `uniform_name`. 

I also think it is fine to have the string template right here in the function because it will not be used anywhere else.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+    return json_parsed
+
+
+def get_shader_from_job(shader_job: str) -> str:
+    """
+    Helper function to get the filename of a shader file from its corresponding shader job.
+    :param shader_job: the path of the shader job file
+    :return: the path of the fragment shader file.
+    """
+    return shader_job.split('.')[0] + '.frag'
```
paulthomson: This could fail if the filename is e.g. 'my.shader.job.json'. I think we use a function called `remove_end` for this; feel free to import it from runspv.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def get_shader_test_from_job(shader_job: str) -> str:
+    """
+    Helper function to get the filename of the new shader test from the given shader job.
+    :param shader_job: the path of the shader job file
+    :return: the path of the shader_test file
+    """
+    return shader_job.split('.')[0] + '.shader_test'
```
paulthomson: Same as above

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
paulthomson: By the way, the Python (3) approach is probably to use pathlib.Path (and then the `.stem` attribute to get the path without the extension) . But Python has a weird mix of usually expecting paths as strings, and sometimes allowing the use of Path. Since you did not use Path from the beginning, it is probably not worth introducing it now.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
paulthomson: But see below, as I have another thought.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: Oops!

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: I had no idea this was a thing - thanks.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: For some reason I thought when writing this that periods were only allowed as the file ending suffix in paths, so that's why I went with that. Thanks for catching this.

PR #578, opened 2019-06-25
Add script to convert a directory of glsl-fuzz shaders into a form suitable for use with spirv-fuzz.

afd:
In the process refactored runspv a bit in order to reuse parts of it.

### DIFF
--- a/python/src/main/python/drivers/runspv.py
+++ b/python/src/main/python/drivers/runspv.py
@@ -327,13 +327,18 @@ def remove_end(str_in: str, str_end: str):
     return str_in[:-len(str_end)]
 
 
-# Runs spirv-opt on |spv_file| with arguments |spirv_opt_args|.  The optimized binary file is
-# |output|.
 def run_spirv_opt(
     spv_file: str,
     spirv_opt_args: List[str],
     output: str
 ) -> None:
+    """
+    Optimizes a SPIR-V file.
+
+    :param spv_file: name of SPIR-V file to be optimized
+    :param spirv_opt_args: arguments to be passed to spirv-opt
+    :param output: name of file into which optimized SPIR-V will be written
+    """
 
     log('Running optimizer.')
 
--- a/python/src/main/python/drivers/runspv.py
+++ b/python/src/main/python/drivers/runspv.py
@@ -338,6 +338,7 @@ def run_spirv_opt(
     :param spv_file: name of SPIR-V file to be optimized
     :param spirv_opt_args: arguments to be passed to spirv-opt
     :param output: name of file into which optimized SPIR-V will be written
+    :return: None
     """
 
     log('Running optimizer.')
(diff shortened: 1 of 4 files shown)


python/src/main/python/drivers/runspv.py
```diff
@@ -327,21 +327,37 @@ def remove_end(str_in: str, str_end: str):
     return str_in[:-len(str_end)]
 
 
+# Runs spirv-opt on |spv_file| with arguments |spirv_opt_args|.  The optimized binary file is
```
paulthomson: It is more common to use docstrings: """ """ within the function body.

python/src/main/python/drivers/shader_job_uniforms_to_spirv_fuzz_facts.py
```diff
@@ -79,9 +83,9 @@ def main() -> None:
                 constantWord=[int_representation])
             fact_list.append(dict(constantUniformFact=fact_constant_uniform))
 
-    with open(os.path.splitext(args.shader_job)[0] + ".facts", "w") as f:
+    with open(args.output_file, "w") as f:
```
paulthomson: You didn't really change this, but it is best to use `encoding='utf-8', errors='ignore'`, or just use `runspv.open_helper`

python/src/main/python/drivers/runspv.py
afd: Done.

python/src/main/python/drivers/shader_job_uniforms_to_spirv_fuzz_facts.py
afd: Done (the latter)

python/src/main/python/drivers/runspv.py
```diff
@@ -329,19 +329,40 @@ def remove_end(str_in: str, str_end: str):
...
 
-    cmd = [spirvopt_path(), spv_file, '-o', result]
+    cmd = [spirvopt_path(), spv_file, '-o', output]
     cmd += spirv_opt_args
 
     subprocess_helper(cmd, timeout=TIMEOUT_SPIRV_OPT_SECONDS)
 
-    return result
+
+# Runs glslangValidator on |glsl_shader| to convert it to SPIR-V.  The resulting SPIR-V file is
```
paulthomson: Here as well.

python/src/main/python/drivers/shader_job_uniforms_to_spirv_fuzz_facts.py
```diff
@@ -16,24 +16,28 @@
...
 import json
-import os
 import struct
+import sys
 from typing import Any, Dict, List
 
 import runspv
 
 
 # Turns a GraphicsFuzz .json file into a spirv-fuzz .facts file, with one fact per word of uniform
```
paulthomson: Could fix this as well

python/src/main/python/drivers/runspv.py
afd: Done.

PR #599, opened 2019-07-09
Fix issue where the reducer was replacing a dead code injection with its body

### LINKED ISSUE #598: Reducer is replacing if(_GLF_DEAD(...)) { } statements with their bodies.
The reducer is wrongly turning:#

if(_GLF_DEAD(...)) { body }

into:

{ body }

even when --preserve-semantics is enabled.

afd:
The reducer used to be very conservative about replacing dead code
injections with their bodies and was made less conservative.  However
this introduced a bug whereby the reducer would always replace
'if(_GLF_DEAD(...)) body' by 'body'.  The problem was that the guard
of the dead code injection was being regarded as itself being dead
code, which is wrong; it should only be the 'then' and 'else' branches
of the if statement that are regarded as dead.

This change fixes this problem.  It was necessary to change a number
of tests in the process: tests where the reducer was only managing to
do a particularly good job of reducing a shader due to this bug; for
example it could turn:

if (_GLF_DEAD(...)) {
  int a;
  a = 2;
}

into:

{
  int a;
  a = 2;
}

which is OK, because the body of the 'if' statement does not have
side-effects to data declared outside the 'if' statement.  But this
was only working due to the body of the 'if' being unconditionally
used to replace the 'if'.  Now that that problem is fixed, the side
effect analysis is not yet sophisticated enough to determine that the
body of the 'if' is side-effect free, thus the transformation does not
take place.

Fixes #598.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
@@ -43,24 +43,26 @@ public class SideEffectChecker {
 
       @Override
       public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
-        if (!TyperHelper.getBuiltins(shadingLanguageVersion, shaderKind)
+        if (TyperHelper.getBuiltins(shadingLanguageVersion, shaderKind)
             .containsKey(functionCallExpr.getCallee())) {
-          // Assume that any call to a non-builtin might have a side-effect.
-          predicateHolds();
-        }
-        for (FunctionPrototype p :
-            TyperHelper.getBuiltins(shadingLanguageVersion, shaderKind)
-                .get(functionCallExpr.getCallee())) {
-          // We check each argument of the built-in's prototypes to see if they require lvalues -
-          // if so, they can cause side effects.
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/graphicsfuzz/common/util/SideEffectChecker.java
```diff
@@ -43,24 +43,26 @@ private static boolean isSideEffectFreeVisitor(IAstNode node,
 
       @Override
       public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
-        if (!TyperHelper.getBuiltins(shadingLanguageVersion, shaderKind)
+        if (TyperHelper.getBuiltins(shadingLanguageVersion, shaderKind)
```
paulthomson: So this is where the bug was?

afd: No, the bug was in ReductionOpportunitiesBase, and was to do with when entering and leaving dead code injections was being tracked.

The fix led to some test failures, some of which meant changing the tests (as explained in the PR description). But a few could be made to pass again by having this refinement to the side effect checker.

Issue #543, opened 2019-06-11
Write new versions of GraphicsFuzz reference shaders that use Integer Functions


afd:
For each shader, foo.frag, in shaders/src/main/glsl/samples/310es, produce a version, foo_integer_functions.frag that is semantically equivalent to foo.frag but that makes significant use of Integer Functions.

In addition, foo_integer_functions.frag may make use of control flow that is available in GLSL ES 3.10 but not GLSL ES 1.00, for example while loops, do...while loops, switch statements, and for loops with empty header components.

abelbriggs1:
Fixed by #605

PR #605, opened 2019-07-11
Add new versions of GraphicsFuzz shaders that use Integer Functions

abelbriggs1:
Implements #543. 

This PR adds modified versions of the default GraphicsFuzz shaders to the 310es directory that use GLES features not found in GLES 1.00, including do/while loops, while loops and Integer Functions.

Note - I couldn't make too many useful changes to colorgrid_modulo and squares - they heavily make use of floating point rather than integers and some parts of them rely on having a for loop(continue statement) rather than another type of control flow. Please make suggestions if you have any ideas for spicing those two up.

### DIFF
--- a/shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
+++ b/shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
@@ -57,11 +57,11 @@ void main()
     } while(i < findMSB(512));
     if(gl_FragCoord.x < resolution.x / 2.0)
     {
-        _GLF_color = vec4(data[0] / 10.0, data[5] / 10.0, data[9] / 10.0, 1.0);
+        _GLF_color = vec4(data[findMSB(0)] / 10.0, data[findLSB(32)] / 10.0, data[findMSB(512)] / 10.0, 1.0);
     }
     else
     {
-        _GLF_color = vec4(data[5] / 10.0, data[9] / 10.0, data[0] / 10.0, 1.0);
+        _GLF_color = vec4(data[findLSB(32)] / 10.0, data[findMSB(512)] / 10.0, data[findMSB(0)] / 10.0, 1.0);
     }
 }
 
--- a/shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
+++ b/shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
@@ -17,6 +17,7 @@
  */
 
 precision highp float;
+precision highp int;
 
 layout(location = 0) out vec4 _GLF_color;
 
@@ -30,16 +31,18 @@ bool checkSwap(float a, float b)
 }
 void main()
 {
+    int msb10 = 1024;
+    int msb9 = 512;
     uint uselessOutVariable;
     float data[10];
-    for(int i = bitfieldReverse(int(injectionSwitch.x)); i < findMSB(1024); i++)
+    for(int i = bitfieldReverse(int(injectionSwitch.x)); i < findMSB(msb10); i++)
     {
         data[i] = float(usubBorrow(uint(10), uint(i), uselessOutVariable)) * injectionSwitch.y;
     }
     int i = bitfieldExtract(int(injectionSwitch.x), bitCount(0), int(injectionSwitch.x));
     do
     {
-        for(int j = bitfieldExtract(int(injectionSwitch.x), 0, 0); j < findLSB(1024); j++)
+        for(int j = bitfieldExtract(int(injectionSwitch.x), 0, 0); j < findLSB(msb10); j++)
         {
             if(uint(j) < uaddCarry(uint(i), 1u, uselessOutVariable))
             {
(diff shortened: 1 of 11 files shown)


shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
```diff
@@ -0,0 +1,67 @@
...
+                float temp = data[i];
+                data[i] = data[j];
+                data[j] = temp;
+            }
+        }
+        i++;
+    } while(i < findMSB(512));
+    if(gl_FragCoord.x < resolution.x / 2.0)
+    {
+        _GLF_color = vec4(data[findMSB(0)] / 10.0, data[findLSB(32)] / 10.0, data[findMSB(512)] / 10.0, 1.0);
```
paulthomson: Perhaps in this branch, you could complicate things a bit more. E.g.

* Make one of the arguments to a built in be an expression that includes a uniform. 

```
findMSB(int(injectionSwitch.y * 32.0))

```

* Make one of the arguments be a variable; the variable could be set depending on a uniform:

```glsl
int temp = 1;
if (injectionSwitch.x < 100.0) {
  temp = 2;
}

// use temp in an argument

```

shaders/src/main/glsl/samples/310es/colorgrid_modulo_intfunctions.frag
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+precision highp float;
+
+layout(location = 0) out vec4 _GLF_color;
+
+uniform vec2 resolution;
+
```
paulthomson: It looks like result of most the built-ins here will be evaluated at compile-time. In particular, `bitfieldInsert(800, 0, 0, 0)` is a cool function, but that will probably just be trivially replaced with `800` at compile-time.

We could try passing a loop counter (`i` or `j`) to a built in. I came up with this:

```glsl
    // within the loop body:
    int another_i = i;
    // Set the 16th bit.
    another_i = bitfieldInsert(another_i, 1, 16, 1);
    int sixteenthBit = int(0x010000 * int(injectionSwitch.y));
    // Unset the 16th bit
    another_i = another_i & ~sixteenthBit;
    // now use another_i instead of i within the loop...but still increment i!
```

This can't be simplified completely at compile-time. We could just modify `i` directly, since we don't actually change its value in the end, but it might be best not to do that so that the compiler can still easily see that the loop iterates 800 times, and so it might be able to do some extra optimizations (although 800 iterations is probably too many for the compiler to bother trying to unroll the loop). 

Of course, doing something like this *everywhere* might complicate things too much but, in general, passing literals to built-ins will probably just get optimized away (but it is still good to do it sometimes).

shaders/src/main/glsl/samples/310es/mandelbrot_blurry_intfunctions.frag
```diff
@@ -0,0 +1,70 @@
...
+  return vec3(float(i) / 50.0, float(i) / 120.0, float(i) / 140.0);
+}
+
+vec3 mand(float xCoord, float yCoord) {
+  float height = resolution.y;
+  float width = resolution.x;
+
+  float c_re = 0.8*(xCoord - width/2.0)*4.0/width - 0.4;
+  float c_im = 0.8*(yCoord - height/2.0)*4.0/width;
+  float x = 0.0, y = 0.0;
```
paulthomson: The way built-ins are used below (in this `mand` function) are probably a bit too simple. I would try to complicate one of them to not just take literals.

shaders/src/main/glsl/samples/310es/prefix_sum_intfunctions.frag
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+precision highp float;
+
+layout(location = 0) out vec4 _GLF_color;
+
+uniform vec2 resolution;
+
+void main(void) {
+  uint uselessOutVariable;
```
paulthomson: Again, I think built-ins here just take literals; it would be much better if some of them were more complicated by *somehow* using data that comes from (at least one of) loop counters, uniforms, fragCoord, etc.

shaders/src/main/glsl/samples/310es/squares_intfunctions.frag
```diff
@@ -0,0 +1,139 @@
...
+layout(location = 0) out vec4 _GLF_color;
+
+uniform float time;
+
+uniform vec2 resolution;
+
+float h_r;
+float s_g;
+float b_b;
+
```
paulthomson: Again, I think the uses are very simple here. But perhaps it is fine to leave one shader with *only* simple uses and see what happens!

shaders/src/main/glsl/samples/310es/bubblesort_flag_intfunctions.frag
abelbriggs1: Done.

shaders/src/main/glsl/samples/310es/mandelbrot_blurry_intfunctions.frag
abelbriggs1: I've complicated it with some uniform dependent expressions.

shaders/src/main/glsl/samples/310es/prefix_sum_intfunctions.frag
abelbriggs1: I added injectionSwitch and put it to good use.

shaders/src/main/glsl/samples/310es/colorgrid_modulo_intfunctions.frag
abelbriggs1: Just for fun, I changed it to `bitfieldInsert(800, i, 0, 0)`; I'm not sure if the compiler can optimize that at compile time when it's checked every loop and the loop isn't unrolled.

afd:
@abelbriggs1 can you please investigate why the CI fails for this PR?  Might be that it needs to be rebased on top of dev.

abelbriggs1:
Looks like a test is failing to validate a shader because of this expression:

`(uaddCarry(uint(4 * GLF_live5j), uint(GLF_live5i), GLF_live5uselessOutVariable)) >= 0`

because you can't compare a uint with an int - I'm not sure why the 0 is signed since the generator should have been given 'uint' as a type for generating this expression. I'll do some more debugging soon.

afd:
@abelbriggs1 now that #670 is merged can your rebase this one and try it again?

PR #613, opened 2019-07-15
Fix piglit converter not accepting blank JSON

abelbriggs1:
We don't need to error check the variable - json.load() just returns nothing if the JSON is empty, and errors out by itself if it detects malformed JSON.

### DIFF
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -182,8 +182,6 @@ def get_json_properties(shader_job: str) -> List:
     """
     with gfuzz_common.open_helper(shader_job, 'r') as job:
         json_parsed = json.load(job)
-    if not json_parsed:
-        raise IOError('Malformed shader job file.')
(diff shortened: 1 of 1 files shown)


python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -182,8 +182,6 @@ def get_json_properties(shader_job: str) -> List:
     """
     with gfuzz_common.open_helper(shader_job, 'r') as job:
         json_parsed = json.load(job)
```
afd: By "blank JSON", do you mean an empty file, or a file containing "{}"?  When there are no uniforms we tend to provide the latter, and I believe that should be parsed into an empty dictionary, which your original code should handle just fine.  Are you talking about an actual empty file?  If so, what's the use case for that?

afd: OK, thanks to @paulthomson I have understood what's going on here - an empty dictionary evaluates to false, so you're to have removed the check.

PR #614, opened 2019-07-15
Improve RNG; use commons-rng

paulthomson:
* 64 bit seed. 
* Use commons-rng ISAAC random number generator. 
* Use rng.spawnChild() wherever possible instead of generating a seed. 
* Pass a rng down instead of a seed where possible. 
* Let the seed argument be a String and parse it as an unsigned 64 bit long integer. 
* Output the seed as an unsigned long. 
* Update some tests that were dependent on having the same RNG. 
* Update some Apache Java packages. 
* Add missing 0 seeds in tests.


generator/src/test/java/com/graphicsfuzz/generator/transformation/DonateLiveCodeTransformationTest.java
```diff
@@ -407,7 +409,10 @@ public void testArrayAccessesAreInBounds() throws Exception {
           GenerationParams.normal(ShaderKind.FRAGMENT, true)
       );
```
afd: Can you add a comment here to explain the logic around failCount?  I guess you must have found that when you upped the seed limit, there were some cases where live code injection was failing (i.e., no code was injected), and this just checks that it doesn't fail too often?  It would be worth remarking that it is theoretically possible for this test to fail due to bad luck.

paulthomson: Done

PR #617, opened 2019-07-16
Add piglit shader_runner_gles3 worker script

### LINKED ISSUE #582: Write script to act as a piglit shader_runner worker for shader jobs
For Mesa testing, it'd be nice to have a worker that used the piglit shader_runner framework to render images and dump them. Thanks to PR #577, we now have a script that converts GraphicsFuzz shader jobs to shader_test files - this will come in handy.

A PNG can be dumped from piglit_framework_gl binaries by passing the argument -png; additionally, adding -auto or -fbo to the arguments will cause the shader_test not to display to the user, and simply dump the PNG/display the test results in the command line. With a little bit of effort, we can probably set up piglit to act as a worker.

abelbriggs1:
Fixes #582.

To run the worker, you need glxinfo and shader_runner_gles3 to be on your PATH - however, the script won't work until my corresponding PR to piglit for a -ignore-missing-uniforms argument is merged (should be soon).

The script is kind of a mess in terms of code quality right now, as my sole goal was 'get it working and usable' - code review is highly appreciated.

### DIFF
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -62,10 +62,11 @@ UNIFORM_TYPES = {
 UNIFORM_DEC = 'uniform'
 
 
-def make_shader_test_string(shader_job: str) -> str:
+def make_shader_test_string(shader_job: str, nodraw: bool) -> str:
     """
     Makes a piglit shader_test from a shader job and shader.
     :param shader_job: The path to the shader job file.
+    :param nodraw: determines if the draw command is added to draw the shader.
     :return: the shader_test
     """
     shader_job_json_parsed = get_json_properties(shader_job)
@@ -81,7 +82,7 @@ def make_shader_test_string(shader_job: str) -> str:
     shader_test_string += make_require_header(shader_version_header) + '\n'
     shader_test_string += make_vertex_shader_header() + '\n'
     shader_test_string += make_fragment_shader_header(shader_file_string) + '\n'
-    shader_test_string += make_test_header(shader_job_json_parsed)
+    shader_test_string += make_test_header(shader_job_json_parsed, nodraw)
 
     return shader_test_string
 
@@ -135,10 +136,11 @@ def make_fragment_shader_header(fragment_shader: str) -> str:
     return frag_header
 
 
-def make_test_header(shader_job_json_parsed: dict) -> str:
+def make_test_header(shader_job_json_parsed: dict, nodraw: bool) -> str:
     """
     Creates the [test] header. Loads uniforms based on the uniforms found in the JSON file.
     :param shader_job_json_parsed: the parsed JSON properties.
+    :param nodraw: determines if the draw command is added to draw the shader.
     :return: the shader_test test header string.
     """
     test_header = TEST_HEADER + '\n'
@@ -148,7 +150,8 @@ def make_test_header(shader_job_json_parsed: dict) -> str:
             uniform_name=uniform_name,
             args=' '.join([str(arg) for arg in value['args']])
         )
-    test_header += DRAW_COMMAND
+    if not nodraw:
+        test_header += DRAW_COMMAND
     return test_header
 
 
@@ -218,10 +221,16 @@ def main_helper(args: List[str]) -> None:
     argparser.add_argument(
         'shader_job', 
         help='Path to the GraphicsFuzz shader job JSON file.')
+
+    argparser.add_argument(
+        '--nodraw',
+        action='store_true',
+        help='Do not draw the shader output when running the test. Useful for crash testing.'
+    )
     
     args = argparser.parse_args(args)
   
-    test_string = make_shader_test_string(args.shader_job)
+    test_string = make_shader_test_string(args.shader_job, args.nodraw)
 
     with gfuzz_common.open_helper(get_shader_test_from_job(args.shader_job), 'w') as shader_test:
         shader_test.write(test_string)
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -152,8 +152,8 @@ def make_test_header(shader_job_json_parsed: dict, nodraw: bool) -> str:
             uniform_name=uniform_name,
             args=' '.join([str(arg) for arg in value['args']])
         )
-    test_header += CLEAR_COMMAND + '\n'
     if not nodraw:
+        test_header += CLEAR_COMMAND + '\n'
         test_header += DRAW_COMMAND
     return test_header
 
@@ -179,12 +179,12 @@ def is_version_header(line: str) -> bool:
     return SHADER_VERSION_FLAG in line
 
 
-def get_json_properties(shader_job: str) -> List:
+def get_json_properties(shader_job: str) -> dict:
     """
-    Helper function to parse a shader job JSON file into a list of properties.
+    Helper function to parse a shader job JSON file into a dict of properties.
     Throws IOError if the file can't be parsed.
     :param shader_job: the path to the shader job file.
-    :return: a list of JSON properties.
+    :return: a dict of JSON properties.
     """
     with gfuzz_common.open_helper(shader_job, 'r') as job:
         json_parsed = json.load(job)
--- a/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
+++ b/python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
@@ -34,6 +34,8 @@ TEST_HEADER = '[test]'
 
 # Draw command for piglit to draw the shader's output. Required in the test header.
 DRAW_COMMAND = 'draw rect -1 -1 2 2'
+# Command to clear the screen to black before drawing, for the test header.
+CLEAR_COMMAND = 'clear color 0.0 0.0 0.0 1.0'
 
 # Strings used to specify the GL version to use in a piglit test's
 # [require] header.
@@ -75,7 +77,7 @@ def make_shader_test_string(shader_job: str, nodraw: bool) -> str:
         shader_file_string = shader.read()
 
     shader_lines = shader_file_string.split('\n')
-    shader_test_string = str('')
+    shader_test_string = ''
     # The version header always has to be on the first line of the shader.
     shader_version_header = shader_lines[0]
 
@@ -150,6 +152,7 @@ def make_test_header(shader_job_json_parsed: dict, nodraw: bool) -> str:
             uniform_name=uniform_name,
             args=' '.join([str(arg) for arg in value['args']])
         )
+    test_header += CLEAR_COMMAND + '\n'
     if not nodraw:
         test_header += DRAW_COMMAND
     return test_header
--- a/python/src/main/python/drivers/piglit-worker.py
+++ b/python/src/main/python/drivers/piglit-worker.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 The GraphicsFuzz Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import shutil
+import argparse
+
+
+# From glsl-to-spv-worker
+def remove(f):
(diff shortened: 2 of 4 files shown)


python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+SHADERTEST_SUFFIX = '.shader_test'
+PNG_FILENAME = 'shader_runner_gles3000.png'
+LOGFILE_NAME = 'piglit_log.txt'
+STATUS_FILENAME = 'STATUS'
+
+GLXINFO_CMD = ['glxinfo', '-B']
+SHADERRUNNER_CMD = ['shader_runner_gles3']
+SHADERRUNNER_ARG_PNG = '-png'
+SHADERRUNNER_ARG_AUTO = '-auto'
+SHADERRUNNER_ARG_UNIFORMS = '-ignore-missing-uniforms'
```
paulthomson: Nit: I think most Python code heavily leans towards splitting words with `_`. `RETURNCODE` is a bit of an exception :-) This could be applied in a few places. I will try to point to most of them.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+
+def run_shader_test(shader_test_file: str, skip_render: bool):
+    """
+    Runs a shader_test file and logs the output. If the shader runner errors out, the error is
+    logged and then raised to the caller.
+    :param shader_test_file: the shader_test file to run.
+    :param skip_render: whether to skip rendering or not.
+    """
+    shader_runner_cmd = SHADERRUNNER_CMD + [shader_test_file, SHADERRUNNER_ARG_AUTO,
+                                            SHADERRUNNER_ARG_UNIFORMS]
```
paulthomson: Please add `-report-subtests` as this causes shader_runner_gles3 to render off-screen.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+    if not skip_render:
+        shader_runner_cmd.append(SHADERRUNNER_ARG_PNG)
+    try:
+        log('Exec: ' + shader_test_file + ' with ' + SHADERRUNNER_CMD[0])
+        results = subprocess.run(shader_runner_cmd, timeout=TIMEOUT, check=True)
```
paulthomson: Use `ruspv.subprocess_helper` with the `verbose=True`. This will ensure the log file contains the output from shader_runner.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+    logged and then raised to the caller.
+    :param shader_test_file: the shader_test file to run.
+    :param skip_render: whether to skip rendering or not.
+    """
+    shader_runner_cmd = SHADERRUNNER_CMD + [shader_test_file, SHADERRUNNER_ARG_AUTO,
+                                            SHADERRUNNER_ARG_UNIFORMS]
+    if not skip_render:
+        shader_runner_cmd.append(SHADERRUNNER_ARG_PNG)
+    try:
+        log('Exec: ' + shader_test_file + ' with ' + SHADERRUNNER_CMD[0])
```
paulthomson: If using `runspv.subprocess_helper` (as advised below), you can probably remove this, as it logs exactly what is executed.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+            log(STDOUT_STR + ex.stdout.decode(encoding='utf-8', errors='ignore'))
+        if ex.stderr is not None:
+            log(STDERR_STR + ex.stderr.decode(encoding='utf-8', errors='ignore'))
+        log(RETURNCODE_STR + str(ex.returncode))
+        raise ex
+    if results.stdout is not None:
+        log(STDOUT_STR + results.stdout.decode(encoding='utf-8', errors='ignore'))
+    if results.stderr is not None:
+        log(STDERR_STR + results.stderr.decode(encoding='utf-8', errors='ignore'))
+    log(RETURNCODE_STR + str(results.returncode))
```
paulthomson: Hopefully most of the above can be removed when you switch to using `runspv.subprocess_helper`

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+    try:
+        run_shader_test(shader_test_file, skip_render)
+    except subprocess.TimeoutExpired:
+        status = STATUS_TIMEOUT
+    except subprocess.CalledProcessError:
+        status = STATUS_CRASH
+
+    # Piglit throws the output PNG render into whatever the current working directory is
+    # (and there's no way to specify a location to write to) - we need to move it to wherever our
+    # output is.
```
paulthomson: This is not ideal as running multiple workers from the same directory could conflict. But I suppose the `worker_info.json` is also an issue (and this is true for the `glsl-to-spv-worker` as well). So maybe we can just accept that workers need to be run in different directories. @afd FYI

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+
+    time.sleep(2)
```
paulthomson: Can this be removed? It is worrying if not! :-)

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+
+    if os.path.isfile(os.getcwd() + '/' + PNG_FILENAME):
+        shutil.move(os.getcwd() + '/' + PNG_FILENAME, os.getcwd() + '/' + png_file)
```
paulthomson: Please remove `os.getcwd()` and `/`. You should use `os.path.join('a', 'b', 'c')` to create a path, but you don't need to here anyway; you can just use `PNG_FILENAME` and `png_file`.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -81,7 +82,7 @@ def make_shader_test_string(shader_job: str) -> str:
     shader_test_string += make_require_header(shader_version_header) + '\n'
     shader_test_string += make_vertex_shader_header() + '\n'
     shader_test_string += make_fragment_shader_header(shader_file_string) + '\n'
-    shader_test_string += make_test_header(shader_job_json_parsed)
+    shader_test_string += make_test_header(shader_job_json_parsed, nodraw)
```
paulthomson: I suggest adding the `clear color 0.0 0.0 0.0 1.0` command, even though it probably won't matter, unless shader_runner definitely does that anyway.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+FRAG_SUFFIX = '.frag'
+JSON_SUFFIX = '.json'
+PNG_SUFFIX = '.png'
+SHADERTEST_SUFFIX = '.shader_test'
+PNG_FILENAME = 'shader_runner_gles3000.png'
+LOGFILE_NAME = 'piglit_log.txt'
+STATUS_FILENAME = 'STATUS'
+
+GLXINFO_CMD = ['glxinfo', '-B']
+SHADERRUNNER_CMD = ['shader_runner_gles3']
```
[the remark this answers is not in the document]
paulthomson: You may want to replace these variables with functions that return the list. If you do this, then I personally think it is fine (better) to have the string literals in the function body (as opposed to adding *more* variables).
E.g.

```py
def glxinfo_cmd():
  return [tool_on_path('glxinfo'), '-B']
```

python/src/main/python/drivers/piglit-worker.py
paulthomson: `nodraw` is an example.

paulthomson: OK, actually, that is all!

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Oops - that was just to test something!

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: That's probably a good idea - just for sanity.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Used `os.getcwd()` because I thought it would resolve an unrelated error when debugging this - thanks for catching.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: It's possible that I could submit a patch upstream to allow specification of a directory argument, but the -png argument actually comes from piglit-gl-framework and not just shader_runner (unlike -ignore-missing-uniforms), which means that I'd be messing with a whole bunch of applications we're not interested in.

python/src/main/python/drivers/piglit-worker.py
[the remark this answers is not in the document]
abelbriggs1: Done.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+    :param skip_render: whether to skip rendering or not.
+    """
+
+    assert os.path.isdir(output_dir)
+    assert os.path.isfile(frag_file)
+    assert os.path.isfile(json_file)
+
+    # graphicsfuzz_piglit_converter has to skip the first argument since it's meant to be a
+    # standalone script taking arguments from a command line.
+    arglist = list('')
```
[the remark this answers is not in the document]
abelbriggs1: You're right, I got a little confused between the two converter files.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: I was inconsistent between splitting words before - I've changed the variables to be split now.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+            if job.noJob is not None:
+                log("No job")
+            elif job.skipJob is not None:
+                log("Skip job")
+                service.jobDone(worker, job)
+            else:
+                assert job.imageJob is not None
+                if job.imageJob.computeSource:
+                    log("Got a compute job, but Piglit doesn't support compute shaders.")
+                    job.imageJob.result.status = tt.JobStatus.UNEXPECTED_ERROR
```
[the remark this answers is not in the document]
abelbriggs1: Oops - fixed.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -0,0 +1,418 @@
...
+        return None, None
+
+
+def dump_glxinfo(filename: str) -> None:
+    """
+    Helper function that dumps the stable results of 'glxinfo -B' to a JSON file. Removes any
+    file with the same name as filename before writing.
+    :param filename: the filename to write to.
+    """
+    check_working_glxinfo()
```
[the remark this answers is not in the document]
abelbriggs1: `dump_glxinfo()` is already called in a try clause, so I suppose that there's no reason to check for errors prematurely.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: This seems to fix the issue where shader_runner renders garbage whenever _GLF_color is not set in a fragment - thanks!

paulthomson:
`graphicsfuzz_piglit_converter.py` `get_json_properties` is annotated as returning a `List` but I think it returns a `Dict`

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -148,7 +152,9 @@ def make_test_header(shader_job_json_parsed: dict) -> str:
             uniform_name=uniform_name,
             args=' '.join([str(arg) for arg in value['args']])
         )
-    test_header += DRAW_COMMAND
+    test_header += CLEAR_COMMAND + '\n'
```
paulthomson: I suppose this should be move down so it is under `if not nodraw`.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
```diff
@@ -62,10 +64,11 @@
 UNIFORM_DEC = 'uniform'
```
paulthomson: `get_json_properties` is annotated as returning a `List` but I think it returns a `Dict`

abelbriggs1: Missed this when I was fixing it up for the original PR - whoops. Funnily enough, the JSON was already annotated correctly in the function that it was used for.

python/src/main/python/drivers/graphicsfuzz_piglit_converter.py
abelbriggs1: Right - we wouldn't need to clear the screen if we're not going to draw to it in the first place.

PR #622, opened 2019-07-17
Add reducer opportunities

jiradeto:
Related issue: #519 

The already implemented reducer that replaces the global variable declarations with the expression is not yet integrated into GraphicsFuzz. This PR adds the implementation for setting up such reducer with the reduction driver and reduction opportunities.


jiradeto:
Seems like the auto-formatting in my editor indented with a wrong number of spaces. Fixed now!

PR #625, opened 2019-07-19
WIP: add skeleton classes for the new primitive mutation

jiradeto:
Add skeleton classes for the review


generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+                + value.getType().toString()
+                + "_" + primitiveValueString.replace(".", "_");
+        List<Declaration> declarations = new ArrayList<>(tu.clone().getTopLevelDeclarations());
+        FunctionPrototype functionPrototype = new FunctionPrototype(functionName,
+            new QualifiedType(value.getType(), Arrays.asList()));
+
+        declarations.add(0, new FunctionDefinition(functionPrototype,
+            new BlockStmt(
+                Arrays.asList(
+                    new ReturnStmt(
```
afd: Here, instead of returning a FloatConstantExpr, you should recursively call generateExpr.  But you should do so with a new FactManager that has each of the function parameters with known values as facts.

For example, if you are generating a function with this signature

_GLF_COMPUTE_42(float _GLF_PARAM_12, float _GLF_PARAM_UNKNOWN, float GLF_PARAM_64) {
   return ...
}

then you want to generate the ... with the facts that _GLF_PARAM_12 has the value 12, and _GLF_PARAM_64 has value 64.

The new fact manager should have the same global-scope facts as the current fact manager (but as it's generating a new function, it can't remember any local scope facts, e.g. about local variables).

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+      case 1:
+        // functionFact
+
+        String functionName =
+            "_GLF_COMPUTE_"
+                + value.getType().toString()
+                + "_" + primitiveValueString.replace(".", "_");
+        List<Declaration> declarations = new ArrayList<>(tu.clone().getTopLevelDeclarations());
+        FunctionPrototype functionPrototype = new FunctionPrototype(functionName,
```
afd: Here, you should decide on a random number of parameters for the function, and a random value for each parameter (randomly choosing some of them to have no known value [the Optional case]).

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+        // Store variable fact in FactManager
+        factManager.getVariableFacts().put(declInfo, new VariableFact(variablesDecl, declInfo,
+            value));
+        return new VariableIdentifierExpr(variableName);
+
+
+      case 1:
+        // functionFact
+
+        String functionName =
```
afd: Similarly, this should in due course be handled in a nicer way via a constant for the prefix and functions to create the name.  And IdGenerator should be used to ensure that multiple functions that generate the same value don't have the same name.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+                                  PipelineInfo pipelineInfo,
+                                  FactManager factManager,
+                                  FunctionDefinition functionDefinition,
+                                  Stmt stmt,
+                                  Value value,
+                                  IRandom generator) {
+
+    String primitiveValueString = value.getData().get(0).get().toString();
+
+    switch (generator.nextInt(2)) {
```
afd: In addition to the cases that you have, consider adding the following cases:

- a case that returns a ConstantExpr with the required value - a FloatConstantExpr for a float value, an IntConstantExpr for an int, etc.

- a case that returns a VariableIdentifierExpr using a variable the fact manager knows about (so don't create a new fact, but use an existing fact)

- a case that returns a FunctionCallExpr using a function the fact manager already knows about

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+        return new FunctionCallExpr(functionName, Arrays.asList());
+    }
+
+
+    return null;
+  }
+
+
+  public Map<VariableDeclInfo, VariableFact> getVariableFacts() {
```
afd: Return a Collections.unmodifiableMap(...)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,116 @@
...
+    return variableFacts;
+  }
+
+  public Map<FunctionPrototype, FunctionFact> getFunctionFacts() {
```
afd: Return a Collections.unmodifiableMap(...)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FunctionFact.java
```diff
@@ -0,0 +1,18 @@
+package com.graphicsfuzz.generator.mutateapi;
+
+import com.graphicsfuzz.common.ast.decl.FunctionPrototype;
+import java.util.List;
+import java.util.Optional;
+
+public class FunctionFact {
```
afd: Make the fields private and final.

Document why the values are optional.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+  public ExpressionGenerator(TranslationUnit translationUnit, PipelineInfo pipelineInfo) {
+    this.translationUnit = translationUnit;
+    this.pipelineInfo = pipelineInfo;
+    idGenerator = new IdGenerator();
+  }
+
+  private Expr generateLiteralNumber(Value value) {
+    assert BasicType.allScalarTypes().contains(value.getType());
+
+    if (value.getType() == BasicType.FLOAT) {
```
afd: I don't understand the purpose of "nameFromNumber".  Why not just:

new FloatConstantExpr(value.getData().get().toString()

?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+      return new FloatConstantExpr(nameFromNumber(value.getData().get(0).get()));
+    }
+    if (value.getType() == BasicType.INT) {
+      return new IntConstantExpr(nameFromNumber(value.getData().get(0).get()));
+    }
+    if (value.getType() == BasicType.UINT) {
+      return new UIntConstantExpr(nameFromNumber(value.getData().get(0).get()) + "u");
+    }
```
afd: Make the following declaration final.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+    for (int i = 0; i < value.getType().getNumElements(); i++) {
+      assert value.getType().getElementType() != value.getType();
+      args.add(generateLiteralNumber(
+          new PrimitiveValue(value.getType().getElementType(),
+              Arrays.asList(Optional.of(value.getData().get(i).get()))
+          )));
+    }
+    return new TypeConstructorExpr(value.getType().toString(), args);
+  }
+
```
afd: A JavaDoc comment explaining the purpose of each parameter would be good here.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+  public Expr generateExpr(FactManager factManager,
+                           FunctionDefinition functionDefinition,
+                           Stmt stmt,
+                           Value value,
+                           IRandom generator) {
+    while (true) {
```
afd: 2 is not right here; it should be 4.  I suggest:

final int NUM_WAYS_TO_GENERATE_EXPR = 4;

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+          final Optional<Expr> expr = factManager.getFact(value);
+          if (expr.isPresent()) {
+            return expr.get();
+          } else {
+            continue;
+          }
+          // Generate ConstantExpr (Float, Int)
+        case 3:
+          return generateLiteralNumber(value);
+        default:
```
afd: It would be good to throw a new RuntimeException("some appropriate error message") here, as the switch cases should really cover all cases.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+          return null;
+      }
+    }
+  }
+
```
afd: A comment for this method explaining what it does, and perhaps giving one or two examples, would be great.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+        + value.getType().toString() + "_"
+        + value.getData().get(0).get().toString().replace(".", "_");
+  }
+
+  // TODO: Requires further implementation in the case of the negative number, i.e., -0.75 could be
+  // converted to _GLF_FLOAT_NEGATIVE_0_75.
+  private String nameFromNumber(Number number) {
+    return number.toString();
+  }
+
```
afd: This looks good.  As it is rather intricate, I think a comment explaining what it does would be good.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+        default:
+          return null;
+      }
+    }
+  }
+
+  // TODO: Handle the cases of vectors (vec4, vec3, vec2).
+  private String genVarName(Value value) {
+    return Constants.GLF_PRIMITIVE_VAR + "_"
+        + value.getType().toString() + "_"
```
afd: Shouldn't the "value.getData().get(0).get().toString().replace(...)" stuff be done in 'nameFromNumber'?

Or, actually, how about you get rid of 'nameFromNumber', and override the toString method of 'value' so that it returns a string?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+
+  // TODO: Requires further implementation in the case of the negative number, i.e., -0.75 could be
+  // converted to _GLF_FLOAT_NEGATIVE_0_75.
+  private String nameFromNumber(Number number) {
+    return number.toString();
+  }
+
+  private String getFunctionName(Value value) {
+    return Constants.GLF_COMPUTE + "_"
+        + value.getType().toString().toUpperCase() + "_"
```
afd: Again, I think the "value.getData().get(0).get()).replace(...)" stuff would be better off as a toString method on PrimitiveValue.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+        + nameFromNumber(value.getData().get(0).get()).replace(".", "_")
+        + "_ID_" + idGenerator.freshId();
+  }
+
```
afd: I'm not really sure what this method is for and why we need it.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+  // TODO: Handles the case of boolean value.
+  private Optional<Number> numberFromName(String string) {
+    try {
+      return Optional.of(NumberFormat.getInstance().parse(string));
+    } catch (ParseException parseException) {
+      return Optional.empty();
+    }
+  }
+
```
afd: Even though this is a private method, I think some explanatory JavaDoc would be worthwhile.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+  private final PipelineInfo pipelineInfo;
+  private final IdGenerator idGenerator;
+
+  public ExpressionGenerator(TranslationUnit translationUnit, PipelineInfo pipelineInfo) {
+    this.translationUnit = translationUnit;
+    this.pipelineInfo = pipelineInfo;
+    idGenerator = new IdGenerator();
+  }
+
+  private Expr generateLiteralNumber(Value value) {
```
afd: (I wrote this comment a bit later in my review.)

How about giving the Value interface a method:

Expr toExpr()
?

It could return a ConstantExpr or a TypeConstructorExpr for the value.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  private Expr generateVariableFact(FactManager factManager,
+                                    FunctionDefinition functionDefinition,
+                                    Stmt stmt,
+                                    Value value,
+                                    IRandom generator) {
+    final String variableName = genVarName(value);
+    final VariableDeclInfo declInfo =
+        new VariableDeclInfo(variableName, null,
```
afd: new Initializer(new FloatConstantExpr(...))

should be:

new Initializer(generateExpr(...));

i.e., generate the initial expression recursively.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+
+public class ExpressionGenerator {
+
+  private static final int MAX_FUNCTION_PARAMS = 5;
+  private static final int INT_MIN = -(1 << 17);
+  private static final int INT_MAX = 1 << 17;
+
+  private final TranslationUnit translationUnit;
+  private final PipelineInfo pipelineInfo;
+  private final IdGenerator idGenerator;
```
afd: I think it would be good to have a field:

private final IRandom generator;

and then get rid of "generator" from all the argument lists.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+
```
afd: ^ I suggest having an additional field, "currentDepth", which is private but not final and gets initialized to 0.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
+public class ExpressionGenerator {
+
+  private static final int MAX_FUNCTION_PARAMS = 5;
+  private static final int INT_MIN = -(1 << 17);
+  private static final int INT_MAX = 1 << 17;
```
afd: Have a private static final int MAX_DEPTH = // something reasonable (perhaps 5)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+          )));
+    }
+    return new TypeConstructorExpr(value.getType().toString(), args);
+  }
+
+  public Expr generateExpr(FactManager factManager,
+                           FunctionDefinition functionDefinition,
+                           Stmt stmt,
+                           Value value,
+                           IRandom generator) {
```
afd: I suggest you structure this as follows:

if (currentDepth >= MAX_DEPTH) {
  return a literal expr
}

currentDepth++;
Expr result = null;
while (true) {
   // the code you have now, but assign to result and break from the loop rather than returning.
}
currentDepth--;
return result;

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
afd: Actually you need to use IdGenerator here to give the function an associated id, like you do with variable names.  I suggest calling the function genFunctionName.

The reason you need to use IdGenerator is that you might end up generating several functions to compute e.g. the constant 3.  They can't share a name.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+    return new VariableIdentifierExpr(variableName);
+  }
+
+  private Expr generateFunction(FactManager factManager,
+                                FunctionDefinition functionDefinition,
+                                Stmt stmt,
+                                Value targetValue,
+                                IRandom generator) {
+
+    final String functionName = getFunctionName(targetValue);
```
afd: Explain in comments why a child fact manager is needed.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+    final FactManager childFactManager = factManager.clone();
```
afd: ^ Rather than cloning the fact manager, I suppose we would really like a new fact manager that has global scope facts available to it.

Perhaps give FactManager a method, "getGlobalFacts", which could return this if the fact manager has no prototype, and could return prototype.getGlobalFacts() otherwise.  That should give the facts at global scope.  Then the child fact manager could be a clone of factManager.getGlobalFacts().

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+    } catch (ParseException parseException) {
+      return Optional.empty();
+    }
+  }
+
+  private Expr generateVariableFact(FactManager factManager,
+                                    FunctionDefinition functionDefinition,
+                                    Stmt stmt,
+                                    Value value,
+                                    IRandom generator) {
```
afd: We could generate local variables, or global variables.  The choice could be made randomly here.

A global variable fact could be added to the fact manager's global facts.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+        childFactManager);
+    final FunctionPrototype prototype = new FunctionPrototype(functionName,
+        targetValue.getType(),
+        params);
+    final FunctionDefinition newFunctionDef = new FunctionDefinition(prototype,
+        new BlockStmt(new ArrayList<>(), false)
+    );
+    newFunctionDef.getBody().addStmt(new ReturnStmt(
+        generateExpr(childFactManager, newFunctionDef, stmt, targetValue, generator)));
+    translationUnit.addDeclaration(newFunctionDef);
```
afd: When you add the fact about the function to the fact manager, it should be added as a global fact.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/Value.java
```diff
@@ -0,0 +1,36 @@
...
+import com.graphicsfuzz.common.ast.type.Type;
+import com.graphicsfuzz.generator.fuzzer.Fuzzer;
+import com.graphicsfuzz.generator.semanticschanging.LiteralFuzzer;
+import java.util.List;
+import java.util.Optional;
+
+public interface Value {
+
+  Type getType();
+
```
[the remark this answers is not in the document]
afd: Thinking about this more, can we change "valueIsKnown" to be called "valueIsUnknown", so that it returns true if and only if the value is totally unknown?  (This will require flipping all uses of the method).

My reasoning is that in the case of a composite value, we currently say that "valueIsKnown" is true if there is some list of values available, even if some of the values in the list are themselves unknown or contain unknown values inside them.  So really by saying "valueIsKnown" we mean "we know something about this value", not "it is completely known".

I think changing things so that "valueIsUnknown" returns true if the value is *totally* unknown would be clearer.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+import com.graphicsfuzz.common.ast.expr.Expr;
+import com.graphicsfuzz.common.ast.type.BasicType;
+import com.graphicsfuzz.common.ast.type.Type;
+import com.graphicsfuzz.generator.semanticschanging.LiteralFuzzer;
+import java.util.Optional;
+
+public class BooleanValue implements Value {
+
+  private final Optional<Boolean> value;
+  private final BasicType type;
```
afd: ^ Get rid of the 'type' field - there is no need for it.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!this.valueIsKnown() && !that.valueIsKnown()) {
+      return true;
+    }
+
+    return this.getValue() == that.getValue();
+  }
+
+  @Override
+  public Type getType() {
```
afd: Just return BasicType.BOOL

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public boolean valueIsKnown() {
+    return value.isPresent();
+  }
+
+  @Override
+  public Expr generateLiteral(LiteralFuzzer literalFuzzer) {
+    if (!valueIsKnown()) {
```
afd: Change "type" to "BasicType.BOOL".

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+public class BooleanValue implements Value {
+
+  private final Optional<Boolean> value;
+  private final BasicType type;
+
+  public BooleanValue(Optional<Boolean> value) {
+    this.value = value;
+    this.type = BasicType.BOOL;
+  }
+
```
afd: In the following, I suggest that you do not introduce the "boolean equals(BooleanValue that)" overload of the equals method.

Instead, I suggest something like:

  @Override
  public boolean equals(Object that) {
    if (this == that) {
      return true;
    }
    if (!(that instanceof BooleanValue)) {
      return false;
    }
    final BooleanValue thatBooleanValue = (BooleanValue) that;
    return this.valueIsKnown() == thatBooleanValue.valueIsKnown() && this.getValue() == that.getValue();
  }

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+      return true;
+    }
+
+    if (!this.valueIsKnown() && !that.valueIsKnown()) {
+      return true;
+    }
+
+    return this.getValue() == that.getValue();
+  }
+
```
afd: Don't forget to override hashCode() - you always need to do this when you override equals.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public Expr generateLiteral(LiteralFuzzer literalFuzzer) {
+    if (!valueIsKnown()) {
+      return literalFuzzer.fuzz(type).orElse(null);
+    }
+    return new BoolConstantExpr(value.get());
+  }
+
+
+  @Override
```
afd: Does toString need to be overridden?  Maybe it is nice to override it for ease of debugging.

You should return something like "unknown_bool" if no value is present.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.graphicsfuzz.generator.mutateapi;
+
```
afd: Check for unused imports and remove them.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
```diff
@@ -0,0 +1,150 @@
...
+        return new UIntConstantExpr(value.get().toString() + "u");
+      }
+    }
+    return literalFuzzer.fuzz(basicType).orElse(null);
+  }
+
+  public Optional<Number> getValue() {
+    return value;
+  }
+
```
afd: I did not yet study the logic of the following method, but I suggest you move it out of this class.  It is very much to do with the expression rewriting that the overall tool does, and feels like it doesn't belong to this class.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
```diff
@@ -0,0 +1,150 @@
...
+    if (getType() == BasicType.INT) {
+      int original = value.get().intValue();
+      int left = generator.nextInt(original + 1);
+      int right = original - left;
+      return new ImmutablePair<>(Optional.of(left), Optional.of(right));
+    }
+
+    throw new RuntimeException("Should be unreachable");
+  }
+
```
afd: Make analogous changes to "equals" as suggested in BooleanValue.

afd: And in the equals method, can't we just return:

   this.basicType == thatNumericValue.basicType && this.value.equals(thatNumericValue.value())

?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // TODO: handle unit case
+
+    return false;
+  }
+
+
+  @Override
+  public String toString() {
```
afd: As with BooleanValue, return something different when the value is not present.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,88 @@
...
+  public Type getType() {
+    return type;
+  }
+
+  @Override
+  public boolean valueIsKnown() {
+    return valueList.isPresent();
+  }
+
+  @Override
```
afd: The equals methods look incomplete.

Please adjust as suggested in BooleanValue to have one equals method that overrides the one from Object, and get rid of the "boolean equals(CompositeValue that)" overload.

Then the logic should be easy: just need to return whether the types are .equals and the values are .equals.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,88 @@
...
+      return false;
+    }
+
+    if (!this.valueIsKnown() && !that.valueIsKnown()) {
+      return true;
+    }
+
+    return false;
+  }
+
```
afd: Be sure to override hashCode.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,88 @@
...
+    if (type instanceof BasicType) {
+      final List<Expr> args = valueList
+          .get()
+          .stream()
+          .map(item -> item.generateLiteral(literalFuzzer))
+          .collect(Collectors.toList());
+      return new TypeConstructorExpr(type.toString(), args);
+    }
+
```
afd: //TODO -> // TODO

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+  private final Map<Value, List<VariableFact>> variableFacts;
+  private final Map<Value, List<FunctionFact>> functionFacts;
+
+  public FactManager(FactManager prototype) {
+    this.prototype = prototype;
+    variableFacts = new HashMap<Value, List<VariableFact>>();
+    functionFacts = new HashMap<Value, List<FunctionFact>>();
+  }
+
+  public FactManager newScope() {
```
afd: Add JavaDoc to explain what this method is for.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+    return new FactManager(this);
+  }
+
```
afd: Add JavaDoc to explain what this method is for.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+  public boolean isGlobalManager() {
```
afd: Add comment to explain that global precisely means "has no prototype"

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+    return prototype == null;
+  }
+
```
afd: Perhaps rename this to "getAllVariableFacts", and add JavaDoc.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+  public Map<Value, List<VariableFact>> getVariableFacts() {
+    return Collections.unmodifiableMap(variableFacts);
+  }
+
```
afd: Add JavaDoc.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+  public Optional<List<VariableFact>> getVariableFacts(Value value) {
```
afd: Better to change return type to List<VariableFact> and then do:

return variableFacts.getOrDefault(value, Collections.emptyList());

You can then change calling code to know that there are no facts if the list is empty.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+    return variableFacts.entrySet()
+        .stream()
+        .filter(item -> item.getKey().equals(value))
+        .map(Map.Entry::getValue)
+        .findFirst();
+  }
+
```
afd: Analogous comments to "getFunctionFacts" as per the above for "getVariableFacts".

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+    if (!functionFacts.containsKey(value)) {
+      return Optional.empty();
+    }
+    return functionFacts.entrySet()
+        .stream()
+        .filter(item -> item.getKey().equals(value))
+        .map(Map.Entry::getValue)
+        .findFirst();
+  }
+
```
afd: JavaDoc.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+  public void addVariableFact(Value value, VariableFact variableFact) {
+    final List<VariableFact> newFacts = variableFacts.getOrDefault(value, new ArrayList<>());
+    newFacts.add(variableFact);
+    variableFacts.put(value, newFacts);
+  }
+
```
afd: JavaDoc.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
afd: But actually I wonder whether this method is needed at all?  What's the reason for needing to know all variable facts?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
afd: But actually I think you should do something like the following:

final List<VariableFact> result = new ArrayList<>();
result.addAll(variableFacts.getOrDefault(value, Collections.emptyList());
if (!isGlobal()) {
  result.addAll(prototpye.getVariableFacts(value));
}

That way you get all the facts available from this manager, including those from its ancestors.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
afd: But I think you should get rid of 'getFunctionFacts()'

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+        .filter(item -> item.getKey().equals(value))
+        .map(Map.Entry::getValue)
+        .findFirst();
+  }
+
+  public Map<Value, List<FunctionFact>> getFunctionFacts() {
+    return Collections.unmodifiableMap(functionFacts);
+  }
+
+  public Optional<List<FunctionFact>> getFunctionFacts(Value value) {
```
afd: Function facts can only exist at global scope, so here I think you want to do:

if (isGlobal()) {
  return ... // do the getOrDefault thing
}
return prototype.getFunctionFacts(value);

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
```diff
@@ -0,0 +1,97 @@
...
+        .findFirst();
+  }
+
+  public void addVariableFact(Value value, VariableFact variableFact) {
+    final List<VariableFact> newFacts = variableFacts.getOrDefault(value, new ArrayList<>());
+    newFacts.add(variableFact);
+    variableFacts.put(value, newFacts);
+  }
+
+  public void addFunctionFact(Value value, FunctionFact functionFact) {
```
afd: Let's only add function facts to global scope.

So:

if(globalScope() {
  // do the addition of the fact
} else {
  prototype.addFunctionFact(...);
}

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,607 @@
...
+    final FunctionFact functionFact =
+        functionFacts.get().get(generator.nextInt(functionFacts.get().size()));
+    final List<Value> argValues = functionFact.getArguments();
+    final List<Expr> args = generateFunctionArgs(factManager, functionDefinition, stmt,
+        argValues);
+    return new FunctionCallExpr(functionFact.getFunctionName(), args);
+  }
+
+  private Expr generateKnownVariableFact(FactManager factManager, Value value) {
+
```
afd: I think you can simplify this a lot:

final List<VariableFact> availableFacts = factManager.getVariableFacts(value);
if (availableFacts.isEmpty()) {
  return null;
}
// pick a random variable fact from the list and return as you do currently.

This avoids the need for "getVariableFacts()" and the filtering out of the ones that don't match on the value.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
jiradeto: I implemented hashCode with the idea of using prime numbers here. I am not quite sure how you do it normally. Please let me know what you think.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/FactManager.java
jiradeto: It was used by the former implementation but there is no need to use it now. Thanks for pointing out.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/Value.java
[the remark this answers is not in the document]
jiradeto: Fixed

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
jiradeto: Fixed

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/BooleanValue.java
afd: This is overkill.

First: there's no point doing "getType().hashCode()" since "getType()" always returns the same thing.

Second: Optional already has its own "hashCode()" method.

So you can just do:

@Override
public int hashCode() {
    return value.hashCode();
}

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,101 @@
...
+  public boolean equals(Object that) {
+    if (this == that) {
+      return true;
+    }
+
+    if (!(that instanceof CompositeValue)) {
+      return false;
+    }
+
+    final CompositeValue thatCompositeValue = (CompositeValue) that;
```
afd: Don't use ==.  That compares object references.  You need to use .equals for both type comparisins and value list comparisons.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,101 @@
...
+    return this.getType() == thatCompositeValue.getType()
```
afd: Get rid of the case of 'that' to CompositeValue, and use 'thatCompositeValue' instead.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
```diff
@@ -0,0 +1,101 @@
...
+        && this.getValueList() == ((CompositeValue) that).getValueList();
+  }
+
+  @Override
+  public int hashCode() {
+    int hashCode = 17;
+
+    hashCode = 37 * hashCode + getType().hashCode();
+
```
afd: Just do:

@Override
public int hashCode() {
    return Objects.hash(type, valueList);
}

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+   * @param value              the value that will be computed by the expression generated by
+   *                           this method.
+   * @return the expression whose value is known by the fact manager.
+   */
+  public Expr generateExpr(FactManager factManager,
+                           FunctionDefinition functionDefinition,
+                           Stmt stmt,
+                           Value value) {
+
+    if (currentDepth > MAX_DEPTH) {
```
afd: "the maximum limitation" -> "the maximum limit"

"we don't want the program to go deeper" -> "we don't want generation to go deeper"

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+    }
+    currentDepth--;
+    return result;
+  }
+
+  /**
+   * A method that provides a value obtained by executing a for loop.
+   *
+   * <p>Given the expected value, this method will declare a new zero-initialized variable and a for
+   * loop statement guarantee to make the variable value equals to the given value. To do so, it
```
afd: Space before "(divisor)"

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+   * will generate a random number(divisor) that will be added to the variable for each iteration.
+   * The binary expression adding the remainder to the return value will also be
+   * injected if there is a left over from a division of original value by the divisor.
+   *
```
afd: "x value" -> "a value x"

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+   * @param stmt               statement in the body of the given function where the generated
+   *                           expression will be inserted before.
+   * @param value              the value that will be computed by the expression generated by
+   *                           this method.
+   * @return a variable identifier to the new variable generated by this method.
+   */
+
+  private Expr generateForLoopValue(FactManager factManager,
+                                    FunctionDefinition functionDefinition,
+                                    Stmt stmt, Value value) {
```
afd: It could be UINT as well - please add a TODO for that.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+                stmt,
+                new NumericValue(BasicType.INT, Optional.of(0)))
+        ));
+    final VariablesDeclaration variablesDecl = new VariablesDeclaration(value.getType(),
+        variableDeclInfo);
+    functionDefinition.getBody().insertBefore(stmt, new DeclarationStmt(variablesDecl));
+
+    // Decide whether the loop should be incremented or decremented for each iteration.
+    final boolean isIncrement = generator.nextBoolean();
+    // If value is unknown, we could generate and return any number.
```
afd: A concern here: if we choose a very large integer we may end up with a loop with masses of iterations.

Can you open an issue for that with a TODO that links to the issue?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,629 @@
...
+    if (remainder > 0) {
+      final Value remainderValue = new NumericValue((BasicType) value.getType(),
+          Optional.of(remainder));
+      functionDefinition.getBody().insertBefore(stmt, new ExprStmt(new BinaryExpr(
+          new VariableIdentifierExpr(varName),
+          generateExpr(factManager, functionDefinition, stmt, remainderValue),
+          BinOp.ADD_ASSIGN
+      )));
+    }
+
```
afd: We should also add this to the fact manager as a known variable fact, right?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/CompositeValue.java
jiradeto: Oops, it was used by mistake. Thanks for pointing!

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
afd: Thanks for opening the issue; please add:

// TODO(github issue URL): (very brief summary of issue)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+    return new VariableIdentifierExpr(varName);
+  }
+
+  private Expr generateAdditionValue(FactManager factManager,
+                                     FunctionDefinition functionDefinition,
+                                     Stmt stmt, Value value) {
+    if (!(value instanceof NumericValue)) {
+      return null;
+    }
+
```
afd: Please add a TODO to also support UINT.

In fact, can you open a GitHub issue for UINT support, and add:

// TODO(github issue URL): (brief summary)

to all the places where UINT support should be added?  (Not one issue per place, but just one "support uint" catch-all issue.)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+    if (value.getType() != BasicType.INT && value.getType() != BasicType.FLOAT) {
+      return null;
+    }
+
+    final Pair<Optional<Number>, Optional<Number>> pair = getPairSum(value);
```
afd: To allow test case reduction, we will want to eventually wrap all things that generate a value in _GLF_IDENTITY(literal value, ...), e.g. here we should do, if we are representing 42 as 20 + 22:

_GLF_IDENTITY(42, _GLF_IDENTITY(20, expr_generating_20), _GLF_IDENTITY(22, expr_generating_22)))

For now can you open a GitHub issue for this?

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+   *
+   * @param factManager        manager class holding the value and its associated expression that
+   *                           guarantees to compute the given value.
+   * @param functionDefinition a function into which the new expression will be injected.
+   * @param stmt               statement in the body of the given function where the generated
+   *                           expression will be inserted before.
+   * @param argValues          values of function arguments from which the new values being
+   *                           generated are derived.
+   * @return a list of function argument expressions.
+   */
```
afd: Can you rename this to "generateArgsForFunctionCall" to make it clearer that these are the expression used to invoke the function (rather than the function's formal parameters)?

afd: I think you could give "stmt" a better name, e.g. "stmtToInsertBefore".

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+        variableDeclInfo);
+    final VariableDeclFact variableDeclFact = new VariableDeclFact(variablesDecl,
+        variableDeclInfo, value);
+    if (atGlobalScope) {
+      translationUnit.addDeclarationBefore(variablesDecl, currentFunction);
+      globalFactManager.addVariableFact(value, variableDeclFact);
+    } else {
+      currentFunction.getBody().insertBefore(stmt, new DeclarationStmt(variablesDecl));
+      factManager.addVariableFact(value, variableDeclFact);
+    }
```
afd: Remove excess whitespace (one line is enough)

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+
+
+    return new VariableIdentifierExpr(varName);
+  }
+
```
afd: Please add JavaDoc and give 'stmt' a better name.

In general please add JavaDoc for all the 'generate' methods, even though they are not public - they're going to be so important.  And use whatever name you choose to replace 'stmt' with consistently.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/ExpressionGenerator.java
```diff
@@ -0,0 +1,632 @@
...
+  /**
+   * This function returns a pair of two numbers whose sum is equal to the given value.
+   *
+   * @param value the original value that will be split into two numbers.
+   * @return if value is unknown, returns a pair of empty value. Otherwise find and return two
+   *     numbers that will add up to the given value.
+   */
+  public Pair<Optional<Number>, Optional<Number>> getPairSum(Value value) {
```
afd: It would be nice to involve the fact manager here, to bias towards choosing at least one number for which we have a known fact.

E.g., if we have facts about variables or functions that compute 12, 16, 3, 6, 2, and we're trying to generate 100, then it would be quite nice if we have a reasonable change of choosing 12 + 88, or 94 + 6, or whatever.

For now feel free to file a GitHub issue for this and do TODO(issue url) here.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/NumericValue.java
```diff
@@ -0,0 +1,93 @@
...
+    if (!(that instanceof NumericValue)) {
+      return false;
+    }
+    final NumericValue thatNumericValue = (NumericValue) that;
+    return this.basicType.equals(thatNumericValue.basicType)
+        && this.value.equals(thatNumericValue.getValue());
+  }
+
+  @Override
+  public int hashCode() {
```
afd: Actually using Object.hash to hash together both type and value would be better.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/VariableFact.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.graphicsfuzz.generator.mutateapi;
+
+public abstract class VariableFact {
+  private final Value value;
+
```
afd: Add JavaDoc

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/VariableFact.java
```diff
@@ -0,0 +1,33 @@
...
+  VariableFact(Value value) {
+    this.value = value;
+  }
+
```
afd: JavaDoc

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/VariableFact.java
```diff
@@ -0,0 +1,33 @@
...
+  public Value getValue() {
+    return value;
+  }
+
```
afd: JavaDoc.

generator/src/main/java/com/graphicsfuzz/generator/mutateapi/VariableFact.java
```diff
@@ -0,0 +1,33 @@
...
+  public abstract String getVariableName();
+
```
afd: Remove excess space.

generator/src/main/java/com/graphicsfuzz/generator/tool/ShaderGenerator.java
```diff
@@ -0,0 +1,132 @@
...
+import com.graphicsfuzz.generator.mutateapi.ExpressionGenerator;
+import com.graphicsfuzz.generator.mutateapi.FactManager;
+import com.graphicsfuzz.generator.mutateapi.NumericValue;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Optional;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+
```
afd: Can you adapt this class so that it allows specifying a desired RGBA list of values on the command-line, parses them into floats, and then uses them in the colour assignment?

generator/src/main/java/com/graphicsfuzz/generator/tool/ShaderGenerator.java
```diff
@@ -0,0 +1,132 @@
...
+      ParseTimeoutException, InterruptedException, GlslParserException {
+
+    final TranslationUnit tu = new TranslationUnit(Optional.of(ShadingLanguageVersion.GLSL_410),
+        Arrays.asList(
+            new PrecisionDeclaration("precision mediump float;"),
+            new FunctionDefinition(
+                new FunctionPrototype("main", VoidType.VOID, Collections.emptyList()),
+                new BlockStmt(new ArrayList<>(), false))));
+
+    final PipelineInfo pipelineInfo = new PipelineInfo();
```
afd: Please add seed command line argument - see what the other GraphicsFuzz tools do here.

generator/src/main/java/com/graphicsfuzz/generator/tool/ShaderGenerator.java
```diff
@@ -0,0 +1,132 @@
...
+                null
+            )
+        )
+    ));
+    tu.getMainFunction().getBody().addStmt(colorAssignment);
+    final FactManager mainFactManager = globalFactManager.newScope();
+    final Expr rvalue = expressionGenerator.generateExpr(
+        mainFactManager,
+        tu.getMainFunction(),
+        colorAssignment,
```
afd: Please replace all four of 0.0, 1.0, 0.3, 1.0 with things that can come from the command line.

afd:
Some additional things to mention now related to this PR, in addition to the changes above:

(1) Rename "ShaderGenerator" to "KnownValueShaderGenerator"

(2) Make a new package, com.graphicsfuzz.generator.knownvaluegeneration, and move all of your new files that are currently in com.graphicsfuzz.generator.mutateapi into this package.

(3) Add test classes, BooleanValueTest, NumericValueTest, CompositeValueTest, FactManagerTest, FunctionFactTest, VariableFactTest, with a reasonable effort to add positive and negative tests for the public APIs of those classes.

(4) Rebase on top of dev and ensure that the CI passes.

jiradeto:
Since this PR was initially intended to add class skeletons for the new shader generator. I believe it might be better if we add and implement the associated test classes in a separate PR? I think that this PR is becoming very large now. Please let me know what you think.

afd:
@jiradeto - yes to doing tests in a separate PR - that sounds good.

PR #632, opened 2019-07-23
Piglit worker improvements

### LINKED ISSUE #628: piglit_worker: support STATUS_NONDET
Now that the piglit worker produces fairly consistent images with the GLES worker, we should figure out how to support non-determinant images for this worker. This will probably involve image comparison of some sort and rendering the shader multiple times.

### LINKED ISSUE #629: piglit_worker: use catchsegv to get crash logs/segfault dumps
When shader_runner crashes, we don't get any kind of log back right now - this means that crashes are difficult or impossible to accurate reproduce. We should probably use `runspv`'s `catchsegv` function instead of `subprocess_helper` to get the log.

abelbriggs1:
Fixes #628 and #629. I was going to separate these into two PRs, but since Paul is gone this week it'd be better to not have merge issues later.

### DIFF
--- a/python/src/main/python/drivers/piglit-worker.py
+++ b/python/src/main/python/drivers/piglit-worker.py
@@ -20,11 +20,9 @@ import os
 import sys
 import subprocess
 import shutil
-from subprocess import CalledProcessError
 import time
 
 import gfuzz_common
-import runspv
 import graphicsfuzz_piglit_converter
 
 HERE = os.path.abspath(__file__)
@@ -60,10 +58,6 @@ WORKER_INFO_FILE = 'worker_info.json'
 
 NO_DRAW_ARG = '--nodraw'
 
-RETURNCODE_STR = 'Returncode: '
-STDOUT_STR = 'STDOUT: '
-STDERR_STR = 'STDERR: '
-
 STATUS_SUCCESS = 'SUCCESS'
 STATUS_CRASH = 'CRASH'
 STATUS_TIMEOUT = 'TIMEOUT'
@@ -74,12 +68,16 @@ STATUS_NONDET = 'NONDET'
 TIMEOUT = 30
 
 
-def glxinfo_cmd():
+def glxinfo_cmd() -> list:
     return [gfuzz_common.tool_on_path('glxinfo'), '-B']
 
 
-def shader_runner_cmd():
+def shader_runner_cmd() -> list:
     return [gfuzz_common.tool_on_path('shader_runner_gles3')]
+
+
+def catchsegv_cmd() -> list:
+    return [gfuzz_common.tool_on_path('catchsegv')]
 
 
 def thrift_connect(server: str, worker_name: str, worker_info: str) -> (FuzzerService, str):
@@ -102,18 +100,18 @@ def thrift_connect(server: str, worker_name: str, worker_info: str) -> (FuzzerSe
         transport.open()
 
         # Get worker name
-        runspv.log("Call getWorkerName()")
+        gfuzz_common.log("Call getWorkerName()")
         worker_res = service.getWorkerName(worker_info, worker_name)
         assert type(worker_res) is not None
 
         if worker_res.workerName is None:
             # noinspection PyProtectedMember
-            runspv.log('Worker error: ' + tt.WorkerNameError._VALUES_TO_NAMES[worker_res.error])
+            gfuzz_common.log('Worker error: ' + tt.WorkerNameError._VALUES_TO_NAMES[worker_res.error])
             exit(1)
 
         worker = worker_res.workerName
 
-        runspv.log("Got worker: " + worker)
+        gfuzz_common.log("Got worker: " + worker)
         assert (worker == worker_name)
 
         return service, worker
@@ -133,7 +131,7 @@ def dump_glxinfo(filename: str) -> None:
     # into JSON.
     glxinfo_lines = filter(
         lambda glx_line: 'OpenGL' in glx_line,
-        runspv.subprocess_helper(glxinfo_cmd()).stdout.split('\n'))
+        gfuzz_common.subprocess_helper(glxinfo_cmd()).stdout.split('\n'))
     # We form keys out of the OpenGL info descriptors and values out of the hardware dependent
     # strings. For example, "OpenGL version string: 4.6.0 NVIDIA 430.14" would become
     # { "OpenGL version string": "4.6.0 NVIDIA 430.14" }.
@@ -184,15 +182,15 @@ def do_image_job(image_job: tt.ImageJob, work_dir: str) -> tt.ImageJobResult:
 
     with gfuzz_common.open_helper(log_file, 'w') as f:
         try:
-            runspv.log_to_file = f
+            gfuzz_common.set_logfile(f)
             run_image_job(frag_file, json_file, status_file, png_file,
                           output_dir, image_job.skipRender)
         except Exception as ex:
-            runspv.log(str(ex))
-            runspv.log('Removing status file and continuing...')
+            gfuzz_common.log(str(ex))
+            gfuzz_common.log('Removing status file and continuing...')
             gfuzz_common.remove(status_file)
         finally:
-            runspv.log_to_file = None
+            gfuzz_common.unset_logfile()
 
     if os.path.isfile(log_file):
         with gfuzz_common.open_helper(log_file, 'r') as f:
@@ -248,27 +246,22 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
     shader_test_file = graphicsfuzz_piglit_converter.get_shader_test_from_job(json_file)
 
     try:
-        runspv.log('Creating shader_test file...')
+        gfuzz_common.log('Creating shader_test file...')
         graphicsfuzz_piglit_converter.main_helper(arglist)
     except Exception as ex:
-        runspv.log('Could not create shader_test from the given job.')
+        gfuzz_common.log('Could not create shader_test from the given job.')
         raise ex
 
-    status = STATUS_SUCCESS
+    shader_runner_cmd_list = catchsegv_cmd() + shader_runner_cmd() + \
+        [shader_test_file, SHADER_RUNNER_ARG_AUTO,
+         SHADER_RUNNER_ARG_UNIFORMS, SHADER_RUNNER_ARG_SUBTESTS]
 
-    shader_runner_cmd_list = shader_runner_cmd() + [shader_test_file, SHADER_RUNNER_ARG_AUTO,
-                                                    SHADER_RUNNER_ARG_UNIFORMS,
-                                                    SHADER_RUNNER_ARG_SUBTESTS]
     if not skip_render:
         shader_runner_cmd_list.append(SHADER_RUNNER_ARG_PNG)
-    try:
-        runspv.subprocess_helper(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
-    except subprocess.TimeoutExpired:
-        status = STATUS_TIMEOUT
-    except subprocess.CalledProcessError:
-        status = STATUS_CRASH
 
-    runspv.log('STATUS: ' + status)
+    status = gfuzz_common.run_catchsegv(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
+
+    gfuzz_common.log('STATUS: ' + status)
 
     # Piglit throws the output PNG render into whatever the current working directory is
     # (and there's no way to specify a location to write to) - we need to move it to wherever our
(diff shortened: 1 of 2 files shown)


python/src/main/python/drivers/piglit-worker.py
```diff
@@ -305,32 +320,33 @@ def main():
 
     gfuzz_common.tool_on_path('shader_runner_gles3')
     gfuzz_common.tool_on_path('glxinfo')
+    gfuzz_common.tool_on_path('catchsegv')
```
paulthomson: `catchsegv` (and `glxinfo` actually) are Linux only. It would be ideal for this script to work (in theory) on Mac, Linux, and Windows. I.e. if `glxinfo` is not found, that is fine (which I believe you pretty much already have, apart from the above) and if `catchsegv` is not found, then it should not be used and `gfuzz_common.run` should be used in that case also. However, perhaps this should be addressed in a follow-up PR.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
     # output is.
 
-    if os.path.isfile(PNG_FILENAME):
-        shutil.move(PNG_FILENAME, png_file)
+    if not skip_render and status == STATUS_SUCCESS:
+        # An image was rendered, so we need to check for nondet. We do this by renaming the
+        # rendered image, rendering a second image, and using filecmp to compare the files.
+        assert os.path.isfile(PNG_FILENAME)
+        os.rename(PNG_FILENAME, COMPARE_PNG_FILENAME)
+        status = gfuzz_common.run_catchsegv(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
```
paulthomson: It would be great to add some logging to this section so that it is clear from reading the logs (in the WebUi) what is going on. I.e. You can explain why we are running shader runner again, and a log statement in each "then" and "else" branch to clarify what has been determined. The log statements can also act as comments, and so you may be able to remove some comments once you have added the log statements.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
         shader_runner_cmd_list.append(SHADER_RUNNER_ARG_PNG)
-    try:
-        runspv.subprocess_helper(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
-    except subprocess.TimeoutExpired:
-        status = STATUS_TIMEOUT
-    except subprocess.CalledProcessError:
-        status = STATUS_CRASH
 
-    runspv.log('STATUS: ' + status)
+    status = gfuzz_common.run_catchsegv(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
```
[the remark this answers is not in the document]
abelbriggs1: Good idea - I figured shutil.move() would handle that for me, but it's good to sanity check just in case.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
     # Piglit throws the output PNG render into whatever the current working directory is
     # (and there's no way to specify a location to write to) - we need to move it to wherever our
     # output is.
 
-    if os.path.isfile(PNG_FILENAME):
-        shutil.move(PNG_FILENAME, png_file)
+    if not skip_render and status == STATUS_SUCCESS:
+        # An image was rendered, so we need to check for nondet. We do this by renaming the
+        # rendered image, rendering a second image, and using filecmp to compare the files.
+        assert os.path.isfile(PNG_FILENAME)
```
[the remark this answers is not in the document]
abelbriggs1: See below.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
+        os.rename(PNG_FILENAME, COMPARE_PNG_FILENAME)
+        status = gfuzz_common.run_catchsegv(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
+        # Something is horribly wrong if shader crashes/timeouts are inconsistent per shader.
+        if not status == STATUS_SUCCESS:
+            raise AssertionError("Shader inconsistently fails - check your graphics drivers?")
```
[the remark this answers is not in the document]
abelbriggs1: Raising any exception in `run_image_job` makes the worker return an UNEXPECTED_ERROR since the status file doesn't exist - I figured AssertionError would be the most descriptive exception type to use here. Is there a different exception type I should use?

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
+        assert os.path.isfile(PNG_FILENAME)
```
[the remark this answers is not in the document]
abelbriggs1: See above.

python/src/main/python/drivers/piglit-worker.py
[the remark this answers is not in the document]
paulthomson: That sounds good. I might be misreading this (on GitHub), but it looks like this is not the case; if the status file missing, nothing happens and the result is not really set.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +254,43 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
+        if filecmp.cmp(PNG_FILENAME, COMPARE_PNG_FILENAME):
+            shutil.move(PNG_FILENAME, png_file)
+        else:
+            status = STATUS_NONDET
+            shutil.move(COMPARE_PNG_FILENAME, os.path.join(output_dir, NONDET0_PNG))
+            shutil.move(PNG_FILENAME, os.path.join(output_dir, NONDET1_PNG))
```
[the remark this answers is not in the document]
abelbriggs1: Done.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Done.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Added a bit more logging to clarify.

python/src/main/python/drivers/piglit-worker.py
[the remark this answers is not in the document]
paulthomson: Never mind! I see it now.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -48,40 +46,45 @@
...
 STATUS_SUCCESS = 'SUCCESS'
 STATUS_CRASH = 'CRASH'
 STATUS_TIMEOUT = 'TIMEOUT'
-STATUS_SANITYERROR = 'SANITY_ERROR'
 STATUS_UNEXPECTED = 'UNEXPECTED_ERROR'
 STATUS_NONDET = 'NONDET'
 
 TIMEOUT = 30
 
+use_catchsegv = True
```
abelbriggs1: Using a global boolean here is kind of hacky, but I think it might be a better option than trying to pass yet another boolean argument all the way down the call stack to run_image_job().

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +256,63 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
     # (and there's no way to specify a location to write to) - we need to move it to wherever our
     # output is.
 
-    if os.path.isfile(PNG_FILENAME):
-        shutil.move(PNG_FILENAME, png_file)
+    if not skip_render and status == STATUS_SUCCESS:
+        # An image was rendered, so we need to check for nondet. We do this by renaming the
+        # rendered image, rendering a second image, and using filecmp to compare the files.
+        if not os.path.isfile(PNG_FILENAME):
+            raise Exception("Shader runner successfully rendered, but no image was dumped?")
```
abelbriggs1: Should I go back to throwing AssertionError here?

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +256,63 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
-
-    status = STATUS_SUCCESS
-
-    shader_runner_cmd_list = shader_runner_cmd() + [shader_test_file, SHADER_RUNNER_ARG_AUTO,
-                                                    SHADER_RUNNER_ARG_UNIFORMS,
-                                                    SHADER_RUNNER_ARG_SUBTESTS]
+    shader_runner_cmd_list = shader_runner_cmd() + \
+        [shader_test_file, SHADER_RUNNER_ARG_AUTO,
+         SHADER_RUNNER_ARG_UNIFORMS, SHADER_RUNNER_ARG_SUBTESTS]
+    if use_catchsegv:
```
paulthomson: I would just define `use_catchsegv` as a local variable at the top of this function. Then do the exception thing (catch `Exception` instead of `BaseException` -- I think that should be fine). If we want to make things more flexible later, we can make `use_catchsegv` a parameter, but I doubt we will ever need to.

python/src/main/python/drivers/piglit-worker.py
paulthomson: Feel free to put these back to `assert` statements, or leave them as they are. But I like the message, so if you change them back to an assert, use `assert cond, "message"`

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -248,34 +256,63 @@ def run_image_job(frag_file: str, json_file: str, status_file: str,
...
+        status = \
+            gfuzz_common.run_catchsegv(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True) \
+            if use_catchsegv else \
+            gfuzz_common.subprocess_helper(shader_runner_cmd_list, timeout=TIMEOUT, verbose=True)
+        # Something is horribly wrong if shader crashes/timeouts are inconsistent per shader.
+        if not status == STATUS_SUCCESS:
+            raise Exception("Shader inconsistently fails - check your graphics drivers?")
+        if not os.path.isfile(PNG_FILENAME):
+            raise Exception("Shader runner successfully rendered, but no image was dumped?")
+        try:
```
paulthomson: Please move the `try` up to the start of this block so that exception above will still cause the cleanup to be triggered.

abelbriggs1: Done

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: I was worried about running tool_on_path for every image job, but it's probably no big deal in the long run.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Done.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -48,40 +46,43 @@
...
 STATUS_TIMEOUT = 'TIMEOUT'
-STATUS_SANITYERROR = 'SANITY_ERROR'
 STATUS_UNEXPECTED = 'UNEXPECTED_ERROR'
 STATUS_NONDET = 'NONDET'
 
 TIMEOUT = 30
 
 
-def glxinfo_cmd():
+def glxinfo_cmd() -> list:
```
paulthomson: Sorry I just noticed this: it should be `List[str]`. Do `from typing import List`.

paulthomson: Same below.

python/src/main/python/drivers/piglit-worker.py
```diff
@@ -224,21 +232,26 @@ def do_image_job(image_job: tt.ImageJob, work_dir: str) -> tt.ImageJobResult:
...
     :param png_file: The PNG file to write to.
     :param output_dir: The directory to use for the job.
     :param skip_render: whether to skip rendering or not.
     """
 
+    use_catchsegv = True
+    # noinspection PyBroadException
+    try:
+        gfuzz_common.tool_on_path('catchsegv')
+    except BaseException:
```
paulthomson: Please use `Exception` instead of `BaseException` (and you can then remove the above # noinspection).

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: Didn't know you could annotate the type that the list contains - thanks.

python/src/main/python/drivers/piglit-worker.py
abelbriggs1: BroadException inspection still happens if I use Exception instead of BaseException, but yeah, good catch.

PR #641, opened 2019-07-29
Refactor shift opaque to use identities instead of fuzzed clamped expressions

### LINKED ISSUE #640: Shift-left-then-shift-right may go wrong if random expression is undefined
We have a nice transformation that does:

e
->
(e << clamp(some-expr, 0, something-small)) >> clamp(some-expr, 0, something-small)

"some-expr" is randomly-generated and side-effect free.

An issue is: what if some-expr is not well-defined?  E.g., what if it contains divisions by zero, or over-shifting?  Even though the two occurrences of "some-expr" are syntactically identical, I don't think we can assume that a shader compiler will treat them as having the same value if that value is undefined.

Possible solutions:

- be less ambitious about choosing a value to shift-left-then-shift-right with - e.g. just pick a random integer in the range 0-8 (or whatever is the right range) and then apply identity to it (getting rid of the random expression generation and clamping).

- have a random expression generation mode that ensures expressions are well-define, a la the "safe math" macros that Csmith uses.

abelbriggs1:
Fixes #640. The fuzzed expression in the clamp will actually be used in execution, so I instead decided to use identity expressions as the shift value, which will always be well-defined and will lead to only marginally less complication than the clamped expression.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -185,14 +185,10 @@ public final class OpaqueExpressionGenerator {
    *     This limits our maximum shift value to 7 bits (more on how we shift an opaque one below).
    * Possibilities for generating an opaque zero include:
    *     Shifting an opaque zero by m bits: (opaque zero) >> n or (opaque zero) << n, where n is
-   *     an expression of a clamped value between 0 and m, inclusive:
-   *     n = clamp(fuzzedexpr, (opaque zero), identity(typeconstructor(m)), and m is an integer
-   *     between 0 and 8, inclusive.
+   *     an integer in [0, 8].
    * Possibilities for generating an opaque zero include:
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -220,62 +216,42 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
     // that, we make sure not to shift our 1 bit out of the integer by limiting our maximum shift
     // to 7 bits.
     final int minBitsForLowpUnsignedInt = minBitsForLowpInt - 1;
-    final int maxValue = generator.nextInt(isZero ? minBitsForLowpInt : minBitsForLowpUnsignedInt);
-    // We pass true as constContext when fuzzing here because the expression will be evaluated,
-    // so we don't want any side effects.
-    final Expr shiftValue = makeClampedFuzzedExpr(type, constContext, depth, fuzzer, maxValue);
```
abelbriggs1: The naming of the various shiftValue variables might be a bit confusing - let me know if you think of better names for them, or if you'd rather me just comment what they're supposed to represent.

PR #642, opened 2019-07-29
Add function to generate opaque zero/one from dot product of two vectors

### LINKED ISSUE #634: Opaque zero / opaque one idea: dot operator
For two vectors, v1 and v2 of the same length:

- If v1[i] != 0 <=> v1[i] = 0 then dot(v1, v2) = 0.  This requires all non-zero components are finite (no infinities, no NaNs).  So we could simplify things to construct vectors that are 0 or 1 everywhere, and never 1 in the same place.

- If v1[i] = v2[i] = 0 for all i different from some single index j, and if v[j] = v[i] = 1, then dot(v1, v2) = 1.

abelbriggs1:
Fixes #634.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -175,10 +175,10 @@ public final class OpaqueExpressionGenerator {
 
   /**
    * Function to generate an opaque zero or one by taking the dot product of two vectors. This
-   * function depends on two rules of the dot product. Let u and v be vectors of the same size,
+   * function depends on two rules of the dot product. Let u and v be vectors of the same width,
    * and let i be a valid index into those vectors.
-   *     If u[i] != 0 <==> v[i] = 0, then dot(u, v) = 0. Example: u = (0, 1), v = (1, 0), then
-   *     dot(u, v) = 1 * 0 + 0 * 1 = 0.
+   *     If u[i] != 0 <==> v[i] = 0 for all i, then dot(u, v) = 0. Example: u = (0, 1), v = (1, 0),
+   *     then dot(u, v) = 1 * 0 + 0 * 1 = 0.
    *     If u[i] = v[i] = 0 for all i except for a single index, j, and if u[j] = v[j] = 1, then
    *     dot(u, v) = 0. Example: u = (0, 1), v = (0, 1), then dot(u, v) = 0 * 0 + 1 * 1 = 1.
    * Because fuzzed expressions are not always well defined, we forgo those in favor of simply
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+   * @return Optional.empty() if an opaque value can't be generated, otherwise an opaque value
+   *     made from the dot product of two vectors.
+   */
+  private Optional<Expr> opaqueZeroOrOneDot(BasicType type, boolean constContext,
+                                            final int depth, Fuzzer fuzzer, boolean isZero) {
+    if (type != BasicType.FLOAT) {
+      return Optional.empty();
+    }
+    // Vector sizes need to be in [2, 4].
+    final int vectorSize = generator.nextInt(3) + 2;
```
abelbriggs1: It'd be better if I didn't use magic numbers here, but I couldn't think of what to name the constants, since they're a little arbitrary. Let me know if I should pursue that.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    final Expr dotProductExpr;
+    if (isZero) {
+      // We're basically producing inverse vectors: u[i] = 0 <==> v[i] != 0.
+      // To do that, we need to store which values will be zero and which will not.
+      final List<Boolean> isZeroDeterminants = new ArrayList<Boolean>();
+      for (int i = 0; i < vectorSize; i++) {
```
abelbriggs1: I don't know if there's some kind of initializer in the STL that lets you initialize a arbitrarily sized list with values - if there is, this would be a place to use it. (Or if there's a better way to do this than a list of booleans, let me know.)

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
afd: The "dot" function is defined on float, as well as vec2--vec4, so we may as well allow this transformation to introduce "dot" expressions on float types.  So I suggest:

final int vectorSize = generator.nextPositiveInt(BasicType.VEC4.getNumElements() + 1);

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
afd: I would simplify the two loops to something like:

for (int i = 0; i < vectorSize; i++) {
  final boolean firstIsZero = generator.nextBoolean();
  firstVectorArgs.add( ..... use firstIsZero .... );
  secondVectorArgs.add( ..... use !firstIsZero .... );

afd: i.e. get rid of isZeroDeterminants.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      }
+      assert firstVectorArgs.size() == vectorSize && secondVectorArgs.size() == vectorSize;
+      dotProductExpr = new FunctionCallExpr("dot",
+          new TypeConstructorExpr(
+              BasicType.makeVectorType(type, vectorSize).toString(), firstVectorArgs),
+          new TypeConstructorExpr(
+              BasicType.makeVectorType(type, vectorSize).toString(), secondVectorArgs));
+    } else {
+      // The two vectors will be exactly the same - all 0 except for one value, which will be 1.
+      final int nonZeroIndex = generator.nextInt(vectorSize);
```
afd: In the following, I suggest you have two lists, "firstVectorArguments" and "secondVectorArguments" (or have an array of two lists if you prefer), and then in the loop add, separately, an opaque zero / opaque one to each of them.  This will mean that while their contents will evaluate to zeros and ones in the same order, the methods by which these opaque values are constructed will be very different.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      final List<Expr> vectorArguments = new ArrayList<Expr>();
+      for (int i = 0; i < vectorSize; i++) {
+        vectorArguments.add(
+            i == nonZeroIndex
+            ? makeOpaqueOne(type, constContext, depth, fuzzer)
+            : makeOpaqueZero(type, constContext, depth, fuzzer));
+      }
+      assert vectorArguments.size() == vectorSize;
```
afd: If you make the above change, then there's no need for the clone.

Furthermore, you could factor out the "dotProductExpr" code from the "then" branch and put it after the loop - you would have to make "firstVectorArgs" and "secondVectorArgs" be declared outside the loop for that to work.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
         depth, fuzzer)));
   }
 
+  /**
+   * Function to generate an opaque zero or one by taking the dot product of two vectors. This
+   * function depends on two rules of the dot product. Let u and v be vectors of the same size,
```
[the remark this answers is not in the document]
abelbriggs1: Thanks for catching - width is a better term. I'll change the variables that use 'size' to use 'width' as well.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -172,6 +173,74 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+   * and let i be a valid index into those vectors.
+   *     If u[i] != 0 <==> v[i] = 0, then dot(u, v) = 0. Example: u = (0, 1), v = (1, 0), then
```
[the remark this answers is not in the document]
abelbriggs1: Oops!

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: I didn't think of using BasicType.VEC4.getNumElements(), that's super self documenting - thanks.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: I don't know why I thought I couldn't inline that - whoops.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java

Issue #545, opened 2019-06-11
Write some new reference shaders that use GLSL ES 3.1 features and interesting built-in functions


afd:
The sample shaders that come with GraphicsFuzz currently are a bit limited.  First, they are essentially limited to GLSL ES 1.00 features (even though they may use a newer shading language version in the #version string).  Second, there just aren't that many of them.

Let's write some more!

afd:
This has been addressed via various PRs.

PR #660, opened 2019-08-08
WIP: new 310es shader

jiradeto:
Related issue #545:

**quicksort_palette**
image_0

- Perform quick sort
- Use a do ... while loop
- Use a switch statement
- Use a struct data type

### DIFF
--- a/shaders/src/main/glsl/samples/310es/quicksort_palette.frag
+++ b/shaders/src/main/glsl/samples/310es/quicksort_palette.frag
@@ -1,0 +1,119 @@
+#version 310 es
+
+/*
+ * Copyright 2019 The GraphicsFuzz Project Authors
+ *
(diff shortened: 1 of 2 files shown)


afd:
I find a reference-variant pair that gives these images:

Reference:
reference

Variant:
variant_002

Even though they look identical to the eye, according to our metrics they are rather different.  I guess it is due to the busy textile-like pattern you have used, which looks very nice but is perhaps a bit too noisy.

Would you be able to try something a bit less detailed on each of the strips?  (A plain colour would be boring, so don't go that far, but maybe something a bit more blocky.)

jiradeto:
@afd with git hash 02fa7d9, I have removed the noisy textures from a shader but keep 4 color palettes:

image_0

I believe this would help making the image diffs easier and more reliable. Please let me know if this shader needs further modification.

PR #666, opened 2019-08-09
Avoid injecting switch case labels into inappropriate places

### LINKED ISSUE #665: Generator donates case outside switch statement
This [Archive.zip](https://github.com/google/graphicsfuzz/files/3484283/Archive.zip) contains the original and the variant shaders generated by a test`testGenerateSmallVulkanShaderFamilyWithReferencesAsDonors` which shows that there is the injected case that appears outside the switch statement. This results in the validation error: 
```
variant_000.frag
ERROR: 0:312: 'case' : cannot appear outside switch statement
ERROR: 0:312: '' : compilation terminated
ERROR: 2 compilation errors.  No code generated.
```

afd:
Adds checks to make sure that we don't inject top-level 'case' or
'default' labels into code, by being aware of switch statements during
donation.  Does allow the insertion of top-level break statements into
switch statements (which was previously disallowed due to switch
statements not being considered, and loops being regarded as the only
acceptable place to add break statements).

Fixes #665.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/util/RemoveImmediateContinueStatements.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/util/RemoveImmediateContinueStatements.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The GraphicsFuzz Project Authors
+ * Copyright 2019 The GraphicsFuzz Project Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


generator/src/main/java/com/graphicsfuzz/generator/util/RemoveImmediateContinueStatements.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 The GraphicsFuzz Project Authors
+ * Copyright 2019 The GraphicsFuzz Project Authors
```
paulthomson: Nit: I would leave the copyright year as 2018.

afd: Fixed.

afd:
@paulthomson I have added tests as you suggested.  Can you re-review please?

PR #670, opened 2019-08-09
Fix making array accesses inbounds with uint index accesses

### LINKED ISSUE #669: MakeArrayAccessesInBounds will generate type mismatches if index is uint
`data[(uaddCarry(uint(4 * GLF_live5j), uint(GLF_live5i), GLF_live5uselessOutVariable)) >= 0 && ...]`

The validator doesn't like that a uint is being compared with an integer.

abelbriggs1:
Fixes #669.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java
@@ -24,6 +24,7 @@ import com.graphicsfuzz.common.ast.expr.Expr;
 import com.graphicsfuzz.common.ast.expr.IntConstantExpr;
 import com.graphicsfuzz.common.ast.expr.ParenExpr;
 import com.graphicsfuzz.common.ast.expr.TernaryExpr;
+import com.graphicsfuzz.common.ast.expr.UIntConstantExpr;
 import com.graphicsfuzz.common.ast.type.ArrayType;
 import com.graphicsfuzz.common.ast.type.BasicType;
 import com.graphicsfuzz.common.ast.type.Type;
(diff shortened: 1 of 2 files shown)


generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java
```diff
@@ -51,20 +52,37 @@ public void visitArrayIndexExpr(ArrayIndexExpr arrayIndexExpr) {
...
+                new ParenExpr(arrayIndexExpr.getIndex().clone()),
+                new UIntConstantExpr(getSize(type).toString() + "u"),
+                BinOp.LT),
+            arrayIndexExpr.getIndex(),
+            new UIntConstantExpr("0u"))
+        );
+      } else {
+        // If indexType ends up being null and the expression is still of type uint, it's better to
+        // generate a statement that will fail to validate rather than crash GraphicsFuzz.
+        arrayIndexExpr.setIndex(new TernaryExpr(
```
abelbriggs1: This is a seemingly minor problem I ran into when developing the tests:

```vec3 data[16];
uint uselessOutVariable;
vec3 stuff = data[uaddCarry(19u, 15u, uselessOutVariable)];```

If you try to make this inbounds *in a test case,* indexType will be null and so you'll hit the `else` case here. However, it seems that fuzzing shaders with expressions like these (specifically the one that was causing PR #605 to fail CI) does not cause this to happen. I am not sure why, but it is probably worth looking into a little.

Comment (afd) [reply generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java]: You seem to have something equivalent to that in the tests for this PR, but I could not see it making indexType turn out to be null.

Can you elaborate a bit more?  We indeed don't really want indexType to be null unless there's some good reason (which I can't think of).

Comment (abelbriggs1) [reply generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java]: Yeah - if you put this uaddCarry call inside of the array access directly(like in the above comment), Typer is unable to find its type and so it will be null. However, making a variable and assigning the function call before using the array access with the variable (like it's done in the test) will work properly.

File: generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBoundsTest.java [hunk 2]
```diff
@@ -83,4 +83,44 @@ public void testMatrixVector2() throws Exception {
...
+        + "}";
+    final TranslationUnit tu = ParseHelper.parse(shader);
+    final Typer typer = new Typer(tu);
+    MakeArrayAccessesInBounds.makeInBounds(tu, typer);
+    assertEquals(PrettyPrinterVisitor.prettyPrintAsString(ParseHelper.parse(expected)),
+        PrettyPrinterVisitor.prettyPrintAsString(tu));
+  }
+
+  @Test
+  public void testUIntFunctionCallReturnIndex() throws Exception {
```
afd: Can you change the shading language version, both before and after, to "#version 310 es"?  This is because uaddCarry does not exist before "300 es".

generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java
afd: The issue there turns out to be due to uaddCarry not being available in GLSL ES 3.00.  Being unknown, it has not been given a type.

I think for this function what I would prefer would be:

      Type indexType = typer.lookupType(arrayIndexExpr.getIndex());
      if (indexType == null) {
          return;
      }
      indexType = indexType.getWithoutQualifiers();

And then in the else branch an assertion that indexType is BasicType.INT.

Ultimately we would like the whole AST to be typed, but at present some things can end up being untyped.  At the moment what the method already does is returns in such cases (see the start of the method), so I think it would be good to follow that style here.

abelbriggs1: Oops, that's kind of a silly mistake on my part. No wonder it worked fine in the shader.

All fixed.

generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBoundsTest.java
abelbriggs1: Addressed.

generator/src/main/java/com/graphicsfuzz/generator/transformation/donation/MakeArrayAccessesInBounds.java
```diff
@@ -87,11 +105,17 @@ private static boolean isArrayVectorOrMatrix(Type type) {
...
+    if (!(index instanceof IntConstantExpr || index instanceof UIntConstantExpr)) {
       return false;
     }
-    Integer indexValue = Integer.parseInt(((IntConstantExpr) index).getValue());
-    return indexValue >= 0 && indexValue < getSize(type);
+    Integer indexValue;
+    if (index instanceof IntConstantExpr) {
+      indexValue = Integer.parseInt(((IntConstantExpr) index).getValue());
+      return indexValue >= 0 && indexValue < getSize(type);
+    } else { // index instanceof UIntConstantExpr
```
afd: Because "getValue" returns something with "u" at the end of it, this would cause a NumberFormatException in the case of an in-bounds uint array access.

[This patch](https://github.com/google/graphicsfuzz/files/3498267/patch.txt) adds a test to expose that problem and uses getNumericValue to overcome it.

abelbriggs1: Applied the patch.

PR #677, opened 2019-08-12
Expression Generator: Add global initializer method

### LINKED ISSUE #652: ExpressionGenerator: global variables cannot be initialized with non constants in GLES ES
From this PR https://github.com/google/graphicsfuzz/pull/625, glslangValidator will give an error when initializing global variables with non-constants for GLSL ES. 

When generating new global variables, we better leave them uninitialized and do the initialization in the first of the main method. 
```
float foo;

void _GLF_init_globals() {
    foo = 1.0;
}
void main() {
     _GLF_init_globals();

}


```

jiradeto:
Fixes #652:

This PR adds a new implementation for generating a new method, `_GLF_init_globals`, that is in charge of initializing the global scope variables. 

```
void _GLF_init_globals();  // function prototype.

float foo;
void main() {
     _GLF_init_globals(); // invokes a function to initialize global vars.
}

void _GLF_init_globals() { // function that assigns values for global vars.
    foo = 1.0;
}
```

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
@@ -17,6 +17,7 @@
 package com.graphicsfuzz.generator.knownvaluegeneration;
 
 import com.graphicsfuzz.common.ast.TranslationUnit;
+import com.graphicsfuzz.common.ast.decl.Declaration;
 import com.graphicsfuzz.common.ast.decl.FunctionDefinition;
 import com.graphicsfuzz.common.ast.decl.FunctionPrototype;
 import com.graphicsfuzz.common.ast.decl.Initializer;
@@ -39,6 +40,7 @@ import com.graphicsfuzz.common.ast.stmt.ReturnStmt;
 import com.graphicsfuzz.common.ast.stmt.Stmt;
 import com.graphicsfuzz.common.ast.type.BasicType;
 import com.graphicsfuzz.common.ast.type.Type;
+import com.graphicsfuzz.common.ast.type.VoidType;
 import com.graphicsfuzz.common.util.IRandom;
 import com.graphicsfuzz.common.util.IdGenerator;
 import com.graphicsfuzz.common.util.PipelineInfo;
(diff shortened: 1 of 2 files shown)


generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -486,19 +488,55 @@ private Expr generateVariableFact(Value value,
...
+        atGlobalScope ? null : new Initializer(initializer));
     final VariablesDeclaration variablesDecl = new VariablesDeclaration(value.getType(),
         variableDeclInfo);
     final VariableDeclFact variableDeclFact = new VariableDeclFact(variablesDecl,
         variableDeclInfo, value);
+
     if (atGlobalScope) {
+      // Searching for the already declared globals initializer's function prototype, if not found,
+      // we have to generate a new prototype, function definition and a function call expression at
+      // the top level of main method to invoke the initializer function.
```
afd: How about you change:

FunctionDefinition initGlobalsFunction

to:

Optional<FunctionDefinition> maybeInitGlobalsFunction

and drop the "orElse(null)".

Then you can do:

FunctionDefinition initGlobalsFunction;
if (maybeInitGlobalsFunction.isPresent()) {
  initGlobalsFunction = maybeInitGlobalsFunction.get();
} else {
  // code to make the new function, assigning it to initGlobalsFunction.
}

generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -486,19 +488,55 @@ private Expr generateVariableFact(Value value,
...
-            value
-        )));
-
+        atGlobalScope ? null : new Initializer(initializer));
     final VariablesDeclaration variablesDecl = new VariablesDeclaration(value.getType(),
         variableDeclInfo);
     final VariableDeclFact variableDeclFact = new VariableDeclFact(variablesDecl,
         variableDeclInfo, value);
+
     if (atGlobalScope) {
```
afd: -> "Search for an existing function for initialization of global variables.  If not found, we have to create one and add a call to the top of 'main' to invoke it."

jiradeto:
Actually, I did implement creating this initializer function in the constructor of the generator but, as you said, my concern is if this tool is later integrated with GraphicsFuzz, we might have to refactor it. So I believe we better keep the creation of this function on demand.

PR #682, opened 2019-08-13
Add function to make opaque zero vec3 from cross product of equal vec3s

### LINKED ISSUE #636: Opaque zero idea: cross product of vec3 with itself
For a vec3 v, cross(v, v) is zero (assuming v contains only finite components).

For an existing side-effect free vector expression e, we could make zero via cross(e, e).  There is a risk that the result might be strange if e contains infinities or NaNs, but if this is an existing expression then the risk of weird stuff happening due to non-finite values would already be there.

abelbriggs1:
Fixes #636.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -574,7 +574,7 @@ public final class OpaqueExpressionGenerator {
   }
 
   /**
-   * Function to create an opaque zero (specifically an opaque zero ved3) by taking the cross
+   * Function to create an opaque zero (specifically an opaque zero vec3) by taking the cross
    * product of two equivalent vec3s. This opaque function relies on the fact that if u == v, where
    * u and v are vec3, then cross(u, v) = (0, 0, 0). These equivalent vec3s are produced by
    * fuzzing random float values and applying different identity functions to them per vector.
@@ -602,8 +602,9 @@ public final class OpaqueExpressionGenerator {
     final List<Expr> secondVec3ConstructorArgs = new ArrayList<Expr>();
     for (int i = 0; i < type.getNumElements(); i++) {
       final Optional<Expr> maybeFuzzedFloatLiteral = litFuzzer.fuzz(type.getElementType());
-      // Something went horribly wrong if we got an Optional.empty(), since our type will always
-      // be a valid literal type to fuzz.
+      // Something went horribly wrong if we got an Optional.empty() - we are guaranteed to obtain a
+      // fuzzed expression, since the element type of a vec3 is a float, and it is always possible
+      // to fuzz a float literal.
       assert maybeFuzzedFloatLiteral.isPresent();
       final Expr fuzzedFloatLiteral = maybeFuzzedFloatLiteral.get();
       // We apply different identities to the literals per vector - the two vectors are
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -569,6 +571,54 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+    // Fuzzer object.
+    final LiteralFuzzer litFuzzer = new LiteralFuzzer(generator);
+    // cross(vec3(firstVec3ConstructorArgs), vec3(secondVec3ConstructorArgs))
+    final List<Expr> firstVec3ConstructorArgs = new ArrayList<Expr>();
+    final List<Expr> secondVec3ConstructorArgs = new ArrayList<Expr>();
+    for (int i = 0; i < type.getNumElements(); i++) {
+      final Optional<Expr> maybeFuzzedFloatLiteral = litFuzzer.fuzz(type.getElementType());
+      // Something went horribly wrong if we got an Optional.empty(), since our type will always
+      // be a valid literal type to fuzz.
+      assert maybeFuzzedFloatLiteral.isPresent();
```
abelbriggs1: Wasn't sure whether to assert or throw an exception here - LiteralFuzzer.fuzz() is perfectly capable of returning Optional.empty() in a runtime context, but it should be impossible here.

afd: Assertion is the right thing here.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -569,6 +571,54 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
         fuzzer)));
   }
 
+  /**
```
afd: "ved3" -> "vec3"

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -569,6 +571,54 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
...
+      return Optional.empty();
+    }
+    // We only want literals, so we need to make our own LiteralFuzzer rather than use the supplied
+    // Fuzzer object.
+    final LiteralFuzzer litFuzzer = new LiteralFuzzer(generator);
+    // cross(vec3(firstVec3ConstructorArgs), vec3(secondVec3ConstructorArgs))
+    final List<Expr> firstVec3ConstructorArgs = new ArrayList<Expr>();
+    final List<Expr> secondVec3ConstructorArgs = new ArrayList<Expr>();
+    for (int i = 0; i < type.getNumElements(); i++) {
+      final Optional<Expr> maybeFuzzedFloatLiteral = litFuzzer.fuzz(type.getElementType());
```
afd: I would change to:

// We are guarenteed to obtain a fuzzed expression, since the element type for a vec3 is float, and it is always possible to fuzz a float literal.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Thanks!

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Fixed. I had trouble trying to phrase this comment, so thanks for adding.

afd:
@abelbriggs1 I don't get why the CI is failing.  Can you please rebase on dev, check that mvn package succeeds locally for you, and force-push if it does?

PR #683, opened 2019-08-13
Add identity to multiply rectangular matrix/vector by identity matrix

### LINKED ISSUE #674: Identity transformation idea: left- or right-multiply vector / rectangular matrix by identity matrix
Left-multiplying a matAxB by the identity matBxB matrix has no effect.  A mat1xB is represented by a vecB.

Right-multiplying a matBxA by the identity matBxB matrix has no effect.  A matBx1 is represented by a vecB.

We should incorporate these identities, particularly since they provide a way to exercise non-square matrices more than we currently do.

abelbriggs1:
Fixes #674.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
@@ -93,6 +93,7 @@ public final class OpaqueExpressionGenerator {
     if (shadingLanguageVersion.supportedTranspose()) {
       expressionIdentities.add(new IdentityDoubleTranspose());
     }
+    expressionIdentities.add(new IdentityMatrixMultIdentity());
 
   }
 
@@ -1366,8 +1367,8 @@ public final class OpaqueExpressionGenerator {
   }
 
   /**
-   * Identity function to transpose a matrix twice. When performed, transforms an expression of a
-   * matrix m -> transpose(identity(transpose(identity(m)))).
+   * Identity transformation to transpose a matrix twice. When performed, transforms an expression
+   * of a matrix m -> transpose(identity(transpose(identity(m)))).
    */
   private class IdentityDoubleTranspose extends AbstractIdentityTransformation {
     private IdentityDoubleTranspose() {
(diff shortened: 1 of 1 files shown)


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1395,4 +1396,50 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
...
+   *     and I is the identity matrix of size n x n, then P * I = P.
+   *     If P is a matrix of size n x m, where n is the number of columns and m the number of rows,
+   *     and I is the identity matrix of size m x m, then I * P = P.
+   *     Note that matrices of size 1 x m or n x 1 are vecm or vecn, respectively.
+   * When performed, transforms an expression of a vector or rectangular matrix P ->
+   * identity(identityMatrix * P) or identity(P * identityMatrix).
+   */
+  private class IdentityMatrixMultIdentity extends AbstractIdentityTransformation {
+    private IdentityMatrixMultIdentity() {
+      // We want all non-square matrices and floating point vectors.
```
afd: ".filter(item -> P(item))" keeps only those elements that *satisfy* P.  Looks like you've written this assuming the opposite.

I suggest you pass the union of BasicType.allNonSquareMatrixTypes and VEC2, VEC3, VEC4.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1395,4 +1396,50 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
...
+      super(BasicType.allNumericTypes().stream()
+          .filter(item -> BasicType.allScalarTypes().contains(item))
+          .filter(item -> BasicType.allIntegerTypes().contains(item))
+          .filter(item -> BasicType.allSquareMatrixTypes().contains(item))
+          .collect(Collectors.toList()), false);
+    }
+
+    @Override
+    public Expr apply(Expr expr, BasicType type, boolean constContext, int depth, Fuzzer fuzzer) {
```
afd: Revise comment about giant lambda constructor if you get rid of the lambda.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Oops, you're exactly right. I just figured out that Streams can be concatenated, so I'll use that instead.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Fixed.

PR #693, opened 2019-08-16
Expression Generator: introduce uniforms 

### LINKED ISSUE #671: Expression generator: introduce uniforms to hold known values
Having some known vaues in uniforms (so that they are unknown to the shader compiler) will help make programs harder to statically optimize away to nothing.

- Add a command line option, --max-uniforms, setting the maximum number of allowed uniforms (default: 10).

- As well as being able to introduce new global and local variables, allow the expression generator to introduce known value uniform variables.  In this case, the type of the variable should be a BasicType (and not BasicType.Bool).

- For now, do not allow matrix uniforms either, but file an issue and a TODO to support them in the future (the reason is that Amber currently has some problems with matrix uniforms).

- A uniform with a known value should have the known value added to PipelineInfo.

jiradeto:
Fixes #671: 

This PR adds a new case for the Expression Generator to produce the given value by injecting a new uniform with no maximum and the generator later randomly picks some of them to be pruned.


generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -163,6 +168,45 @@ public Expr generateExpr(FactManager factManager,
     return result;
   }
 
+  private Expr generateUniformValue(Value value) {
+    if (value.valueIsUnknown()) {
```
afd: Unknown is OK: we can choose a random value for the uniform to be initialized to.

generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -163,6 +168,45 @@ public Expr generateExpr(FactManager factManager,
...
+      return null;
+    }
+    if (!(value instanceof NumericValue)) {
```
afd: Restricting to 'numeric' means that we exclude 'vec' and 'mat' types.

I suggest you check that it's either numeric, or composite with a BasicType as its type.  Then have a special case to exclude matrices as per the issue below.

generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -163,6 +168,45 @@ public Expr generateExpr(FactManager factManager,
...
+    //  matrix uniform once Amber has solved the matrix-related issues.
+    if (value.getType() != BasicType.INT && value.getType() != BasicType.FLOAT) {
+      return null;
+    }
+
+    final String uniformName = Constants.GLF_UNIFORM + "_"
+        + value.getType().toString()
+        + parseNameFromValue(value)
+        + freshId();
+
```
afd: The below should generalize fine to other values too (e.g. vectors)

generator/src/main/java/com/graphicsfuzz/generator/tool/KnownValueShaderGenerator.java
```diff
@@ -152,7 +154,11 @@ public static void mainHelper(String[] args) throws ArgumentParserException, IOE
         new ExprStmt(new BinaryExpr(new VariableIdentifierExpr("_GLF_color"),
             new TypeConstructorExpr("vec4", rgbaExprs), BinOp.ASSIGN)));
 
-    final ShaderJob newShaderJob = new GlslShaderJob(Optional.empty(), new PipelineInfo(), tu);
+    final ShaderJob newShaderJob = new GlslShaderJob(Optional.empty(), pipelineInfo, tu);
+    // We randomly choose some of the uniforms generated by the generator to be pruned.
+    PruneUniforms.prune(newShaderJob, generator.nextInt(pipelineInfo.getNumUniforms()),
```
afd: Can you make --max-uniforms be a command-line option, and only prune uniforms if it is set?  (In which case prune to the given value.)

Also, can you make --generate-uniform-bindings a command line option, and make it do what it does in Generate?  (This is important for Vulkan.)

jiradeto: Ah, I am now understanding how those arguments are related to the pruning. Thanks for your helpful feedback :-)

PR #698, opened 2019-08-16
New 310es shader - Muller's method

abelbriggs1:
Related issue: #545 

**muller**
shader_runner_gles3000

 - Given a uniform vec3 `polynomial` representing the constants of a quadratic function, and a uniform vec3 `initial_xvalues` that represents the xvalues to begin the algorithm with (with `initial_xvalues.y` being the midpoint of `[initial_xvalues.x, initialxvalues.z]`), approximates a root of `polynomial` nearest to `initial_xvalues` by approximating using quadratic functions.
 - Uses a while loop
 - Math-heavy
 - Renders red if the test passes (the method finds the correct root, given a small margin of error), otherwise green

From some fuzz testing, this shouldn't be vulnerable to floating point issues and should give consistent results.

### DIFF
--- a/shaders/src/main/glsl/samples/310es/muller.frag
+++ b/shaders/src/main/glsl/samples/310es/muller.frag
@@ -1,0 +1,76 @@
+#version 310 es
+
+/*
+ * Copyright 2019 The GraphicsFuzz Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


afd:
@jarikomppa could you please take a look at this one for stability (using the same CPU technique you used for other shaders)?  As mentioned in my above comment, it appears to be pretty stable.

jarikomppa:
Results from float and double implementations are equal.

PR #702, opened 2019-08-19
gfauto: fix signature_util plus tests

paulthomson:
Also, add pytest tests and related changes.

### DIFF
--- a/gfauto/.flake8
+++ b/gfauto/.flake8
@@ -15,7 +15,7 @@
 # This is not a TOML file; no quoting.
 
 [flake8]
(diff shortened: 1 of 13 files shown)


paulthomson:
Currently, the interestingness test just uses the signature, and the signature just uses the log. I quite like this "abstraction". However, `no_signature` should nearly always be regarded as a bug, and, in the worst case, the signature could be e.g. "glslang_failure" based on the glslang command exiting with status 0. I.e. at the very least, it should capture which tool failed, even if it can't see any error messages.

PR #704, opened 2019-08-19
Add identity to put data in a larger type and pull it back out

### LINKED ISSUE #675: Identity transformation: put matrix/vector data into a larger type and then pull it out again.
GLSL provides interesting ways to construct matrices/vectors from larger and smaller matrices/vectors.

Let's say that matrix/vector type U is larger than a matrix vector type V if U has at least as many rows and at least as many columns as T, and has more of at least one of these.

Then if e is an expression of the smaller type T, we can use the type constructors T and U to make an equivalent expression:

T(U(v))

The application of U creates a matrix/vector with extra space for the larger type, and the application of T contracts this again.  By applying identities as well (i.e. identity(T(idenitity(U(identity(e)))))) this gives us the opportunity to give various matrix/vector types a good workout.  In particular, it is a nice way to introduce more non-square matrices into a shader.

abelbriggs1:
Fixes #675.


generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1418,6 +1421,91 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
...
+    }
+
+    @Override
+    public Expr apply(Expr expr, BasicType type, boolean constContext, int depth,
+                      Fuzzer fuzzer) {
+      assert !type.isScalar();
+      // Our inner constructor type will be equal or larger in width than our given type. The goal
+      // of the next loop is to populate this list with all types that can fit our given type.
+      final List<BasicType> innerConstructorTypes = new ArrayList<BasicType>();
+      for (BasicType constructorType : BasicType.allBasicTypes().stream().filter(
```
abelbriggs1: This loop is a mess of conditionals, but I couldn't think of a more elegant solution since the conditions are pretty intertwined with each other, and combining if statements with && causes typing issues.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -92,6 +92,9 @@ public OpaqueExpressionGenerator(IRandom generator, GenerationParams generationP
     if (shadingLanguageVersion.supportedMixNonfloatBool()) {
       expressionIdentities.add(new IdentityMixBvec());
     }
+    if (shadingLanguageVersion.supportedNonSquareMatrices()) {
```
afd: Seems limiting to only do this if non-square matrices are supported, since e.g. we can put a mat2x2 into a mat3x3.

It would be better to have the code that applies the identity decide whether to allow square matrices or not.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1418,6 +1421,91 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
     }
   }
 
+  /**
+   * Identity transformation to insert a vector or matrix into a larger vector or matrix using a
+   * type constructor, then extract it back out again using another type constructor. When
```
afd: Replace "->" with "to" (since in some languages "->" can be used as part of a type, and you just said "with type t").

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1418,6 +1421,91 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
...
+   * where m is a type of equal or greater width than t.
+   * The rules for this smaller -> larger -> smaller transformation are as follows:
+   *     If the given type is a vector, the inner constructor type can be a vector of the same
+   *     element type with equal/greater width than the given type, or (provided that the given type
+   *     is a floating point genType) it can be a matrix with equal/greater column width than the
+   *     given type.
+   *     If the given type is a matrix, the inner constructor type can be a matrix with
+   *     equal/greater column/row width than the given type.
+   */
+  private class IdentityCompositeConstructorExpansion extends AbstractIdentityTransformation {
```
afd: Below, and in the above comment, would there be any harm in applying this to scalars too - i.e. pushing a scalar into a bigger type and then pulling it out again?

E.g.:

  float v = float(mat2x4(1.0));

is valid and should give 1.0.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
```diff
@@ -1418,6 +1421,91 @@ public boolean preconditionHolds(Expr expr, BasicType basicType, boolean constCo
...
+    private IdentityCompositeConstructorExpansion() {
+      super(BasicType.allNumericTypes().stream().filter(
+          item -> !item.isScalar()).collect(Collectors.toList()), true);
```
afd: It seems limiting to disallow this for bool and bvec types - we'd like to be able to push a bool into a bvec3, or a bvec3 into a bvec4 - so long as the shading language supports boolean vector types.  So perhaps add bool and bvec types here if the shading language version allows it.

Also: maybe we do not want to allow types that cannot be put into something larger; i.e. not ivec4, uvec4, bvec4 and mat4x4.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
afd: I wonder whether something like this might be cleaner:

if (type is not matrix) {
  for i in [length of vector + 1, 4] {
    add vector type with same element type and base type i
  }
  if element type is float {
    add all relevant matrix types
  }
} else {
  // loops to work out which larger matrices to add
}

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: I did have boolean types enabled when working on this, but enabling them triggers an assertion failure in makeLiteralZeroOrOne():

`assert type.getElementType() != BasicType.BOOL;`

which seems a bit strange. From the context of the calling function (makeOpaqueZeroOrOne()), it seems like if the type is a boolean, then we should never be too deep (and so we should never have to make a literal zero or one) - but I'm not sure why that would be.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: Yeah, you're right! Additionally, I barely even have to change the logic to accommodate scalars.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: The suggested changes to the conditional logic make this much easier to handle inside of the function, so I've implemented this.

generator/src/main/java/com/graphicsfuzz/generator/fuzzer/OpaqueExpressionGenerator.java
abelbriggs1: This is much better - I was trying to have one all-encompassing loop, but breaking down and having multiple loops makes the logic much simpler.

afd:
@abelbriggs1 [This patch](https://github.com/google/graphicsfuzz/files/3686643/patch.txt) adds support for boolean vectors.

Can you please rebase against dev (it has some fixes to boolean vector typechecking that are needed), and then apply the patch and push?

I can then approve and merge.

PR #705, opened 2019-08-20
gfauto: generate amber code for comparing two shader jobs

asuonpaa:
Modified amber generation script to add two pipelines and comparison of their framebuffers when reference shader job is present.

### DIFF
--- a/gfauto/gfauto/amber_converter.py
+++ b/gfauto/gfauto/amber_converter.py
@@ -562,9 +562,9 @@ def graphics_shader_job_amber_test_to_amber_script(
 
     # Add fuzzy compare of framebuffers if there's more than one pipeline
 
-    for pipeline_index in range(len(jobs) - 1):
+    for pipeline_index in range(1, len(jobs)):
         prefix0 = jobs[0].name_prefix
-        prefix1 = jobs[pipeline_index + 1].name_prefix
+        prefix1 = jobs[pipeline_index].name_prefix
         result += f"EXPECT {prefix0}_framebuffer RMSE_BUFFER {prefix1}_framebuffer TOLERANCE 7"
 
     if amberfy_settings.extra_commands:
--- a/gfauto/gfauto/amber_converter.py
+++ b/gfauto/gfauto/amber_converter.py
@@ -521,7 +521,8 @@ def graphics_shader_job_amber_test_to_amber_script(
     jobs = [shader_job_amber_test.variant]
 
     if shader_job_amber_test.reference:
-        jobs.append(shader_job_amber_test.reference)
+        assert isinstance(shader_job_amber_test.reference, GraphicsShaderJob)  # noqa
+        jobs.insert(0, shader_job_amber_test.reference)
 
     for job in jobs:
         prefix = job.name_prefix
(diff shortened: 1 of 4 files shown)


asuonpaa:
@googlebot I signed it!

gfauto/gfauto/amber_converter.py
```diff
@@ -513,51 +513,62 @@ def get_amber_script_shader_def(shader: Shader, name: str) -> str:
...
 
     result = get_amber_script_header(amberfy_settings)
 
-    variant = shader_job_amber_test.variant
+    jobs = [shader_job_amber_test.variant]
 
-    variant_vertex_shader_name = f"{variant.name_prefix}_vertex_shader"
-    variant_fragment_shader_name = f"{variant.name_prefix}_fragment_shader"
+    if shader_job_amber_test.reference:
+        jobs.append(shader_job_amber_test.reference)
```
paulthomson: I just realized: we should probably run the reference first. This is how we have typically done it in the past. So please `.insert` the reference at position 0

paulthomson: Annoyingly, you will need to add: `assert isinstance(shader_job_amber_test.reference, GraphicsShaderJob)  # noqa`

The # noqa is needed because assertions are generally banned from non-test code, but it is only thing the type checker recognizes.

gfauto/gfauto/amber_converter.py
```diff
@@ -513,51 +513,62 @@ def get_amber_script_shader_def(shader: Shader, name: str) -> str:
...
 
-    result += "\n"
-    result += variant.uniform_definitions
+        result += get_amber_script_shader_def(
+            job.fragment_shader, fragment_shader_name
+        )
 
-    # Currently, this buffer must be called "framebuffer" to be able to dump it from Amber.
-    result += "\nBUFFER framebuffer FORMAT B8G8R8A8_UNORM\n"
+        # Define uniforms for variant shader job.
```
paulthomson: Nit: remove "variant" from comment?

gfauto/gfauto/amber_converter.py
```diff
@@ -513,51 +513,62 @@ def get_amber_script_shader_def(shader: Shader, name: str) -> str:
...
+
+        result += f"\nCLEAR {prefix}_pipeline\n"
+        result += f"RUN {prefix}_pipeline DRAW_RECT POS 0 0 SIZE 256 256\n"
+        result += "\n"
+
+    # Add fuzzy compare of framebuffers if there's more than one pipeline
 
-    result += "\nCLEAR gfz_pipeline\n"
-    result += "RUN gfz_pipeline DRAW_RECT POS 0 0 SIZE 256 256\n"
+    for pipeline_index in range(len(jobs) - 1):
```
paulthomson: A bit picky, but I think we would prefer to always compare to the first pipeline (usually the reference, assuming you make the change above). So perhaps do `range(1, len(jobs))` and compare [0] and [pipeline_index]

gfauto/whitelist.dic
```diff
@@ -172,3 +172,6 @@ vec2
 vec4
 framebuffer
 gfautotests
+prefix0
```
paulthomson: One exception to what I said: please prefer splitting words wherever possible (even when ending with a number). I think this is a Python convention, or maybe just something I adopted. So please change to prefix_0, prefix_1, and remove these lines from the whitelist.

paulthomson:
Please search `android_device.py` and `host_device_util.py` for `-i`; wherever this flag is passed to Amber, please add `-I` `variant_framebuffer`. Since the current fuzzes ignores the image, this maybe doesn't matter much, but Amber might complain if it can't find the buffer.

PR #716, opened 2019-08-24
KnownValue Generator: add support for UINT type

### LINKED ISSUE #661: ExpressionGenerator: support type UINT
On #625, the generator is currently unable to generate an expression of UINT. Once that PR gets merged, we should consider supporting type UINT.

jiradeto:
Fixes #661:

This PR adds uint type support for Expression Generator. As well as updated the LiteralFuzzer to provide the fuzzed expression of uint type.


generator/src/main/java/com/graphicsfuzz/generator/knownvaluegeneration/ExpressionGenerator.java
```diff
@@ -561,7 +571,7 @@ private String parseNameFromValue(Value value) {
     if (value instanceof NumericValue) {
       final NumericValue numericValue = (NumericValue) value;
       float floatValue = numericValue.getValue().get().floatValue();
-      if (floatValue < 0.0) {
+      if (String.valueOf(floatValue).contains("-")) {
```
jiradeto: I found that we could end up having a random float -0.0 making the previous version of this condition does not work. Thus, I believe that to know whether the number is negative, checking if the string contains "-" would be nicer.

PR #717, opened 2019-08-28
gfauto: Add support for generating multipipeline amber file from a source directory

### LINKED ISSUE #658: Add support to gfauto for generating wrong image CTS patches
The gfauto tool, which is currently in [this branch](https://github.com/google/graphicsfuzz/tree/gfauto) (due to be merged soon) supports continuous fuzzing using GraphicsFuzz to find crash bugs, de-duplicate them, and transform unique crash bugs into patches that are nearly ready to be added to CTS (requiring a small amount of manual triage).

We would like an analogous flow for wrong image bugs, where GraphicsFuzz identifies a pair of equivalent shaders that lead to radically different images being rendered (due to a shader compiler bug).  We would like to be able to de-duplicate such issues and, for a unique wrong image bug generate a CTS patch.

The focus of this issue is on the second thing: generating a CTS patch.

Specifically, having found a pair of shaders, shader1 and shader2, that are equivalent but are found to expose a wrong image bug, we want to be able to obtain an Amber file that:

- Creates to graphics pipelines, pipeline1 and pipeline2 that use shader1 and shader 2, respectively
- Runs both pipelines
- Asserts that the resulting framebuffers are similar according to a fuzzy diff metric

Amber provides support for running multiple pipelines, and has a root mean square error diff facility thanks to [this PR](https://github.com/google/amber/pull/602).

asuonpaa:
Added support for converting a gfauto source directory containing wrong image case into an amber script which runs all the variants and compares the results against a reference.

Fixes #658

### DIFF
--- a/gfauto/gfauto/amber_converter.py
+++ b/gfauto/gfauto/amber_converter.py
@@ -444,20 +444,24 @@ class ShaderJobFile:
 @attr.dataclass
 class ShaderJobBasedAmberTest:
     reference: Optional[ShaderJob]
-    variant: ShaderJob
+    variants: List[ShaderJob]
 
 
 @attr.dataclass
 class ShaderJobFileBasedAmberTest:
     reference_asm_spirv_job: Optional[ShaderJobFile]
-    variant_asm_spirv_job: ShaderJobFile
+    variants_asm_spirv_job: List[ShaderJobFile]
 
     def to_shader_job_based(self) -> ShaderJobBasedAmberTest:
+        variants = [
+            variant_asm_spirv_job.to_shader_job()
+            for variant_asm_spirv_job in self.variants_asm_spirv_job
+        ]
         return ShaderJobBasedAmberTest(
             self.reference_asm_spirv_job.to_shader_job()
             if self.reference_asm_spirv_job
             else None,
-            self.variant_asm_spirv_job.to_shader_job(),
+            variants,
         )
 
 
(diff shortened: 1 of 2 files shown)


gfauto/gfauto/amber_converter.py
```diff
@@ -513,18 +517,19 @@ def get_amber_script_shader_def(shader: Shader, name: str) -> str:
...
-    assert isinstance(shader_job_amber_test.variant, GraphicsShaderJob)  # noqa
 
     result = get_amber_script_header(amberfy_settings)
 
-    jobs = [shader_job_amber_test.variant]
+    jobs = shader_job_amber_test.variants
 
     if shader_job_amber_test.reference:
         assert isinstance(shader_job_amber_test.reference, GraphicsShaderJob)  # noqa
         jobs.insert(0, shader_job_amber_test.reference)
```
paulthomson: I think this will modify the `shader_job_test` input. Maybe just do `jobs = [shader_job_amber_test.reference] + jobs` or something like that. Or ensure that jobs is a copy when first assigned (which I think you can do with a slice...maybe `jobs = shader_job_amber_test.variants[:]`)

gfauto/gfauto/amber_converter.py
```diff
@@ -633,7 +639,7 @@ def spirv_asm_shader_job_to_amber_script(
 ) -> Path:
 
     log(
-        f"Amberfy: {str(shader_job_file_amber_test.variant_asm_spirv_job.asm_spirv_shader_job_json)} "
+        f"Amberfy: {str(variant.asm_spirv_shader_job_json) for variant in shader_job_file_amber_test.variants_asm_spirv_job} "
```
paulthomson: My IDE doesn't like this, but it works! Unfortunately, I think it prints "<generator object spirv_asm_shader_job_to_amber_script.<locals>.<genexpr> at 0x7fb51f4caaf0>" so please change it anyway! Wrapping it in `[` `]` seems to work for me.

PR #733, opened 2019-09-09
Enhance reducer to include guard of compound statement when flattening it

### LINKED ISSUE #732: reducer: reduce if(func()) -> func()
I believe I saw:

`if (func()) {}`

which could be reduced to:

`func();`

without --preserve-semantics.

afd:
When simplifying a statement of the form:

if(e) {
  s1; s2; ...
}

the reducer used to turn it into just:

{ s1; s2; ... }

It now turns it into:

{ e; s1; s2; ... }

For, while and do-while loops are similarly handled so that their
guard (and the initializer and increment components, in the case of a
for loop) are included in the flattened construct that the reducer
produces.

This is useful in cases where e.g. something appearing in the guard is
needed to trigger a bug, but the control flow statement itself is not
needed.

Fixes #732.


paulthomson:
Just a thought about this: should we instead make it so the "replace if with its body" outputs the guard + the body? Because if "replace if with its body" and "replace if with its guard" both fail, we have increased the number of reduction steps, perhaps quite significantly.

PR #745, opened 2019-09-18
Update generate_cts_test_template.py

ilkkasaa:
Changes the paths to be relative to the python script itself
instead of the working directory.


ilkkasaa:
@googlebot I signed it!

PR #753, opened 2019-09-19
Fix acosh() input range

jarisiru:
acosh() is only defined for input values >=1, but the shader feeds values that are less than that (including negative values). This leads to undefined behavior as per SPIR-V spec. In C++ the result may have bee NaN, which is optional in Vulkan.


paulthomson:
This seems to fix it:

1 -> 1.0
1000 -> 1000.0

jarisiru:
Updated

afd:
@paulthomson I believe Jari addressed your changes.  Can you approve?  I'll then merge.

PR #767, opened 2019-10-03
Zoom into mandelbrot lake to avoid float issues

jarisiru:
Due to the iterative nature of mandelbrot fractal calculations, even minute differences of float accuracy will make the images incomparable. Zooming into the lake will make the images equal while retaining the complexity of the functions.


jarisiru:
(1) It already zooms into 256x256 different places in the lake. If you wish the output image to look like something (rather than a single color), the "lake color" vector could be easily set to xpos/ypos values.
(2) Valid point
(3) Okay.

afd:
Regarding (1): yes, I think doing something based on x and y positions would be good - so long as it's likely to still be numerically stable.

afd:
@jarikomppa [This patch](https://github.com/google/graphicsfuzz/files/3686724/patch.txt) fixes some test code that referred to "mandelbrot_blurry".  Can you apply it please?  Then we should be good to merge.

afd:
Due to the iterative nature of mandelbrot fractal calculations, even
minute differences of float accuracy will make the images
incomparable. Zooming into the lake will make the images equal while
retaining the complexity of the functions.

PR #768, opened 2019-10-03
Add fixed point variants of the Mandelbrot shaders

jarisiru:
Added variants of the fixed point shaders where the actual fractal
calculation is done via fixed point math to avoid floating point issues.
Otherwise the shaders are as before.

For fixed point, 8.8 was chosen so that 32 bit integers are enough in
all situations (fixed point multiplications and divisions require double
the bits of the operands)


jarisiru:
@googlebot I fixed it.

PR #770, opened 2019-10-03
Added fixed point variants of mandelbrot shader

jarisiru:
Added variants of the fixed point shaders where the actual fractal calculation is done via fixed point math to avoid floating point issues.

Otherwise the shaders are as before. For fixed point, 8.8 was chosen so that 32 bit integers are enough in all situations (fixed point multiplications and divisions require double the bits of the operands)


shaders/src/main/glsl/samples/300es/mandelbrot_fixpoint.frag
```diff
@@ -0,0 +1,75 @@
...
+layout(location = 0) out vec4 _GLF_color;
+
+uniform vec2 resolution;
+
+vec3 pickColor(int i) {
+  return vec3(float(i) / 50.0, float(i) / 120.0, float(i) / 140.0);
+}
+
+vec3 mand(float xCoord, float yCoord) 
+{
```
afd: These C-style casts aren't valid in GLSL; you need to do int(...).

shaders/src/main/glsl/samples/310es/mandelbrot_fixpoint.frag
```diff
@@ -0,0 +1,75 @@
...
+layout(location = 0) out vec4 _GLF_color;
+
+uniform vec2 resolution;
+
+vec3 pickColor(int i) {
+  return vec3(float(i) / 50.0, float(i) / 120.0, float(i) / 140.0);
+}
+
+vec3 mand(float xCoord, float yCoord) 
+{
```
afd: Similar.

PR #777, opened 2019-10-21
Add constant folding to array size

jarisiru:
This change adds constant folder in array size fields which
supports most of the operators as well as constant
variables.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
@@ -17,27 +17,48 @@
 package com.graphicsfuzz.common.ast.decl;
 
 import com.graphicsfuzz.common.ast.IAstNode;
+import com.graphicsfuzz.common.ast.expr.Expr;
+import com.graphicsfuzz.common.ast.expr.IntConstantExpr;
 import com.graphicsfuzz.common.ast.visitors.IAstVisitor;
+import com.graphicsfuzz.common.ast.visitors.UnsupportedLanguageFeatureException;
 import java.util.Optional;
 
 public class ArrayInfo implements IAstNode {
 
-  private Optional<Integer> size;
+  private Optional<Expr> size;
+  private Optional<Expr> originalSize;
 
-  public ArrayInfo(Integer size) {
+  public ArrayInfo(Expr size) {
     this.size = Optional.of(size);
+    this.originalSize = Optional.of(size);
   }
 
   public ArrayInfo() {
     this.size = Optional.empty();
+    this.originalSize = Optional.empty();
   }
 
   public boolean hasSize() {
     return size.isPresent();
   }
 
-  public Integer getSize() {
+  public Integer getSize() throws UnsupportedLanguageFeatureException {
+    if (size.get() instanceof IntConstantExpr) {
+      return ((IntConstantExpr)size.get()).getNumericValue();
+    }
+    throw new UnsupportedLanguageFeatureException("Not a constant expression");
+  }
+
+  public Expr getSizeExpr() {
     return size.get();
+  }
+
+  public void setSizeExpr(Expr expr) {
+    this.size = Optional.of(expr);
+  }
+
+  public Expr getOriginalSizeExpr() {
+    return originalSize.get();
   }
 
   @Override
@@ -47,7 +68,7 @@ public class ArrayInfo implements IAstNode {
 
   @Override
   public ArrayInfo clone() {
-    return hasSize() ? new ArrayInfo(getSize()) : new ArrayInfo();
+    return hasSize() ? new ArrayInfo(getSizeExpr()) : new ArrayInfo();
   }
 
   @Override
--- a/ast/src/main/java/com/graphicsfuzz/common/ast/visitors/AstBuilder.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/ast/visitors/AstBuilder.java
@@ -16,6 +16,8 @@
 
 package com.graphicsfuzz.common.ast.visitors;
 
+import com.graphicsfuzz.common.ast.IAstNode;
+import com.graphicsfuzz.common.ast.IParentMap;
 import com.graphicsfuzz.common.ast.TranslationUnit;
 import com.graphicsfuzz.common.ast.decl.ArrayInfo;
 import com.graphicsfuzz.common.ast.decl.Declaration;
@@ -85,6 +87,7 @@ import com.graphicsfuzz.common.ast.type.TypeQualifier;
 import com.graphicsfuzz.common.ast.type.UnknownLayoutQualifier;
 import com.graphicsfuzz.common.ast.type.VoidType;
 import com.graphicsfuzz.common.glslversion.ShadingLanguageVersion;
+import com.graphicsfuzz.common.typing.ScopeTrackingVisitor;
 import com.graphicsfuzz.common.util.ShaderKind;
 import com.graphicsfuzz.parser.GLSLBaseVisitor;
 import com.graphicsfuzz.parser.GLSLParser;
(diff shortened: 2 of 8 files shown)


ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,37 +17,58 @@
 package com.graphicsfuzz.common.ast.decl;
 
 import com.graphicsfuzz.common.ast.IAstNode;
+import com.graphicsfuzz.common.ast.expr.Expr;
+import com.graphicsfuzz.common.ast.expr.IntConstantExpr;
 import com.graphicsfuzz.common.ast.visitors.IAstVisitor;
+import com.graphicsfuzz.common.ast.visitors.UnsupportedLanguageFeatureException;
 import java.util.Optional;
 
 public class ArrayInfo implements IAstNode {
```
afd: Please add a comment to explain the reason for having two size-related fields.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,37 +17,58 @@
...
   }
 
   public ArrayInfo() {
     this.size = Optional.empty();
+    this.originalSize = Optional.empty();
   }
 
   public boolean hasSize() {
     return size.isPresent();
   }
```
afd: Let's rename "getSize" to "getConstantSize" to make it clearer that it is not a plain getter method.

Can you add a JavaDoc comment for this method?  (We try to do that for new code.)

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,37 +17,58 @@
...
+  public Integer getSize() throws UnsupportedLanguageFeatureException {
+    if (size.get() instanceof IntConstantExpr) {
+      return ((IntConstantExpr)size.get()).getNumericValue();
+    }
+    throw new UnsupportedLanguageFeatureException("Not a constant expression");
+  }
+
+  public Expr getSizeExpr() {
     return size.get();
   }
```
afd: Is there ever a use case for setting the size expr to something that is not an IntConstantExpr?  If not then I suggest calling this "setConstantSize" and making it take an IntConstantExpr, with a JavaDoc comment explaining the purpose of it.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,37 +17,58 @@
...
+    return originalSize.get();
+  }
+
   @Override
   public void accept(IAstVisitor visitor) {
     visitor.visitArrayInfo(this);
   }
 
   @Override
   public ArrayInfo clone() {
```
afd: Hmm, I think we probably want to clone the original size expression as well, otherwise cloning will change pretty printing.

Could you make a test class, ArrayInfoTest, and add a test that checks that a shader with a foldable array pretty prints to the same thing after cloning?

ast/src/main/java/com/graphicsfuzz/common/ast/visitors/AstBuilder.java
```diff
@@ -195,9 +198,124 @@ void addTopLevelDeclaration(Declaration decl) {
     topLevelDeclarations.add(decl);
   }
 
+  private static TranslationUnit fixUpArraySizes(TranslationUnit tu) {
```
afd: Please add a brief comment to explain the nature of the patching up, and that it is more convenient to do this after the AST has been built rather than during.

ast/src/main/java/com/graphicsfuzz/common/ast/visitors/AstBuilder.java
```diff
@@ -195,9 +198,124 @@ void addTopLevelDeclaration(Declaration decl) {
...
+              throw new RuntimeException("Unable to fold constant " + bexpr.getText());
+          }
+          return new IntConstantExpr(Integer.toString(fval));
+        }
+        throw new RuntimeException("Unable to fold constant " + expr.getText());
+      }
+
+      @Override
+      public void visitVariableDeclInfo(VariableDeclInfo variableDeclInfo) {
+        if (variableDeclInfo.hasArrayInfo() && variableDeclInfo.getArrayInfo().hasSize()) {
```
afd: I prefer to use final variables where possible in the code, and to use a functional style if possible, so I would prefer just:

variableDeclInfo.getArrayInfo().setSizeExpr(reduce(variableDeclInfo.getArrayInfo().getSizeExpr()));

or:

final Expr constantFoldedSize = reduce(variableDeclInfo.getArrayInfo().getSizeExpr());
variableDeclInfo.getArrayInfo.setSizeExpr(constantFoldedSize);

I also think if might be better to do:

if (varibleDeclInfo.hasArrayInfo()) {
  final ArrayInfo arrayInfo = variableDeclInfo.getArrayInfo();
  ...
}

to avoid looking up variableDeclInfo.getArrayInfo() so many times.

ast/src/main/java/com/graphicsfuzz/common/ast/visitors/AstBuilder.java
```diff
@@ -195,9 +198,124 @@ void addTopLevelDeclaration(Declaration decl) {
...
+          variableDeclInfo.getArrayInfo().setSizeExpr(expr);
+        }
+        super.visitVariableDeclInfo(variableDeclInfo);
+      }
+    }.visit(tu);
+    return tu;
+  }
+
   public static TranslationUnit getTranslationUnit(Translation_unitContext ctx,
                                                    ShaderKind shaderKind, boolean hasWebGlHint) {
```
afd: Can you add a brief comment here to explain why we fix up array sizes?  (Similar to the comment inside 'fixUpArraySizes' but briefer.)

ast/src/test/java/com/graphicsfuzz/common/util/ParseHelperTest.java
```diff
@@ -566,17 +566,61 @@ private String getStringFromInputStream(InputStream strippedIs) throws IOExcepti
   }
 
   @Test
```
afd: Thanks for these tests.  Can you add something to PrettyPrinterVisitorTest to check that an example that uses non-literal array sizes pretty-prints to itself?

Also, do we have any tests in this PR that check the folding does the right thing?  If not it would be good to have them.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
afd: Now that I look at this again, I think it would be better to make constantSize be Optional<Integer> (no need for it to actually be IntConstantExpression), initialize it to Optional.empty(), and then only set it during constant folding.  That seems cleaner than having it temporarily be an arbitrary expression.

For the comments, I would say something like:

"originalSize" is the expression in the original shader representing the array size, and is empty if no size was given.  If "originalSize" is present, constant folding can be used to turn the expression into an integer value, which can be stored in "constantSize".  It is useful to keep "originalSize" around to allow the shader to be pretty-printed in its original form.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,27 +17,95 @@
...
+   */
+  private Optional<Expr> originalSize;
 
-  public ArrayInfo(Integer size) {
-    this.size = Optional.of(size);
+  /**
+   * Constructor, stores the original expression and a clone for constant folding.
+   * @param size Array size expression
+   */
+  public ArrayInfo(Expr size) {
```
afd: I would not clone "size" here - we don't do that elsewhere in the code base, instead preferring to clone when we know we need to rather than by default - and I'd rather stay consistent.

Can you change the parameter name to "originalSize" assuming you make the previous change?

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: Changed constantSize to integer, rename originalSize to just sizeExpr.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: Changed parameter to int.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: Removed cloning from there. For some reason tests pass without clone() anywhere now. Not sure why. They did fail at some point.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: I think I've done all of those.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,27 +17,95 @@
...
   public ArrayInfo() {
-    this.size = Optional.empty();
+    this.constantSize = Optional.empty();
+    this.originalSize = Optional.empty();
   }
 
+  /**
+   * Query for whether this array has a size definition.
+   * @return true if size definition is available
+   */
```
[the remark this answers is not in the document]
jarisiru: In the end I found I need accessor for both hasConstantSize and hasSizeExpr.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,27 +17,95 @@
...
+  }
+
+  /**
+   * Get the constant size.
+   * Assumes that constant folding has occurred before this is called, so calling this while the
+   * AST is being built will likely result in exception.
+   * @return Integer value of the array size
+   * @throws UnsupportedLanguageFeatureException if folding was not done
+   */
+  public Integer getConstantSize() throws UnsupportedLanguageFeatureException {
```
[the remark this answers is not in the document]
jarisiru: Done

jarisiru:
I think I've covered all change requests. Still unsure as to why tests pass even though I have not cloned the expression before reducing.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -47,17 +126,18 @@ public void accept(IAstVisitor visitor) {
 
   @Override
   public ArrayInfo clone() {
-    return hasSize() ? new ArrayInfo(getSize()) : new ArrayInfo();
+    return hasSizeExpr() ? new ArrayInfo(constantSize, sizeExpr) :
```
afd: Even if !hasSizeExpr(), returning new ArrayInfo(constantSize, sizeExpr) will do the right thing.

I thus suggest just returning that, so that it doesn't look like the zero-argument constructor means something special here.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
```diff
@@ -17,27 +17,106 @@
...
+   * Get the constant size.
+   * If constant folding was not performed yet (for example, the AST is stil being built) or if
+   * the array has no size declaration, calling this function will throw an exception.
+   * @return Integer value of the array size
+   * @throws UnsupportedLanguageFeatureException if folding was not done
+   */
+  public Integer getConstantSize() throws UnsupportedLanguageFeatureException {
+    if (hasConstantSize()) {
+      return constantSize.get();
+    }
```
afd: I would like to understand where the array size is required during AST creation.  I wonder whether we can avoid that.

afd:
[This patch](https://github.com/google/graphicsfuzz/files/3782907/patch.txt) adds tests to MakeArrayAccessesInBoundsTest that expose limitations I had not foreseen related to array members in structs.  It also adds disabled tests for analogous issues related to array parameters.

The issue is that struct members that have ArrayInfo objects attached to them are not having folding applied to those ArrayInfo objects.  The patch updates the anonymous class of AstBuilder to add such folding, so that those tests pass.

This allows the check in ArrayInfo in getConstantSize to be removed.

There is still a problem with array parameters, but array parameters need a good overhaul in general, so I have made a low-priority issue (#784) and referenced this issue in various places.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: This was discussed offline.

ast/src/main/java/com/graphicsfuzz/common/ast/decl/ArrayInfo.java
jarisiru: Done

PR #795, opened 2019-11-06
Reducer: simplify literals to canonical form

afd:
Before this change, the reducer would not attempt to simplify
literals, so that e.g. 8092.6 would remain in a reduced shader, even
if in dead code.  This was based on the assumption that the exact form
of final literals would not really be important.  However, having
worked a lot with the reducer it is certainly the case that a final
shader that does not involve weird literal values is easier to
understand than one that does.  This change allows the reducer to
simplify any literal that is not in canonical form to one that is.

We should keep an eye on the knock-on effect this has on reducer
performance; by creating more reduction opportunities we could see a
performance drop.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/CheckAstFeatureVisitor.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/CheckAstFeatureVisitor.java
@@ -73,6 +73,9 @@ public abstract class CheckAstFeatureVisitor extends ScopeTrackingVisitor {
    * Use this method to register that the feature of interest has been found.
    */
   protected void trigger() {
-    this.triggerFunction = Optional.of(getEnclosingFunction());
+    if (!atGlobalScope()) {
+      // Only set the trigger function if we are in some function.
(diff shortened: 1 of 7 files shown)


afd:
@paulthomson I resurrected this PR now that the reducer guards against out-of-bounds accesses during a semantics-changing reduction.

There is still the possibility of an out-of-bounds array access being introduced if:

- live-injected code indexes into an array of size 1
- the expression used for indexing is simplified to the literal 1

This could be solved by changing canonical constants from 1 to 0; I am reluctant to do that due to it introducing a much higher rate of division-by-zero.

I think the likelihood of this being a problem is really small (as we don't use any shaders with arrays of size 1), so I'm inclined to leave it as is and see whether it bites.

Regarding making the better simplification of literals a clean-up pass: that would be a bit complicated, and my hunch is that having it be a regular pass will not add too much overhead, since currently the reducer already tries turning all non-constant expressions into constants.  I suggest we leave it as part of the core "ExprToConstant" pass and keep an eye on reduction times.

afd:
Abandoning due to branch mess - opened #877 instead.

PR #812, opened 2019-12-09
Throw exception on shader version or case difference

jarisiru:
Current version causes unrelated tests to fail (e.g, testGenerateSmallVulkanShaderFamily). Investigating.


jarisiru:
Rebasing to master went bad, I'll have to redo this pull request.

PR #814, opened 2019-12-10
Use reference shaders as donor shaders to avoid version conflicts.

afd:


generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateCodeTransformation.java
```diff
@@ -138,10 +138,28 @@ private TranslationUnit prepareTranslationUnit(File donorFile, IRandom generator
...
+      tu.addDeclaration(new VariablesDeclaration(
+          new QualifiedType(BasicType.VEC4, Collections.singletonList(TypeQualifier.MEDIUMP)),
+          new VariableDeclInfo(addPrefix(OpenGlConstants.GL_FRAG_COLOR), null, null)));
+    }
+    // Remove 'in', 'out' and 'layout' qualifiers from global variables.  This means that, for
+    // instance, the output colour variable of a donated fragment shader will change from:
+    //   layout(location = 0) out vec4 name;
+    // to:
+    //   vec4 name;
+    tu.getTopLevelDeclarations().stream().filter(item -> item instanceof VariablesDeclaration)
```
paulthomson: Maybe this is picky but since the tests failed: I wonder whether an ordinary for loop might be clearer here since you are iterating over the list in order to modify it. Maybe `forEach` is made for this purpose, but streams kind of make me think of just reading items, and maybe forEach would be used to do some action per item, but maybe not modify the items in-place?

PR #817, opened 2019-12-13
Throw exception if shader version or kind doesn't match

jarisiru:


generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateCodeTransformation.java
```diff
@@ -248,11 +248,22 @@ private Stmt prepareStatementToDonate(IInjectionPoint injectionPoint,
...
         // No compatible donors were found, thus we cannot do serious code donation here;
         // we return a null statement instead.
         return new NullStmt();
+      } else {
+        if (shadingLanguageVersion != maybeDonor.get().getShadingLanguageVersion()) {
+          throw new RuntimeException("Incompatible versions: ref=" + shadingLanguageVersion
+              + " donor=" + maybeDonor.get().getShadingLanguageVersion());
+        }
+        if (this.generationParams.getShaderKind() != maybeDonor.get().getShaderKind()) {
+          throw new RuntimeException("Incompatible shader types: ref="
```
afd: Please change "ref" to "reference" in the above.

generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/CheckDonorVersionMatchesTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class CheckDonorVersionMatchesTest {
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test(expected = RuntimeException.class)
```
afd: As "RuntimeException" is a very general class, there is a danger that this test will pass if some unrelated RuntimeException is thrown.

I suggest instead wrapping the code that is expected to throw an exception in a "try ... catch" as follows:

try {
  callToFunction(); // call that should throw a RuntimeException
  fail("An exception should have been thrown.");
} catch (RuntimeException runtimeException) {
  assertTrue(runtimeException.startsWith("Incompatible versions");
}

generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/CheckDonorVersionMatchesTest.java
```diff
@@ -0,0 +1,73 @@
...
+    fileOperations.writeShaderJobFile(es310ShaderJob, reference);
+
+    final GenerationParams normalGenerationParams = GenerationParams.normal(ShaderKind.FRAGMENT, false);
+    final DonateLiveCodeTransformation transformation = new DonateLiveCodeTransformation(
+        TransformationProbabilities.ALWAYS::donateLiveCodeAtStmt,
+        donors,
+        normalGenerationParams,
+        false);
+
+    TranslationUnit tu = es310ShaderJob.getFragmentShader().get();
```
afd: I suppose the following is the call that should be wrapped.

generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateCodeTransformation.java
jarisiru: done

generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/CheckDonorVersionMatchesTest.java
jarisiru: done

generator/src/test/java/com/graphicsfuzz/generator/transformation/donation/CheckDonorVersionMatchesTest.java
jarisiru: done

PR #820, opened 2019-12-16
Teach typechecker about GraphicsFuzz macros

afd:
This change teaches the typechecker what type to give expressions that
are invocations of GraphicsFuzz macros, such as _GLF_DEAD and
_GLF_IDENTITY.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/typing/Typer.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/typing/Typer.java
@@ -103,7 +103,7 @@ public class Typer extends ScopeTrackingVisitor {
   public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
     super.visitFunctionCallExpr(functionCallExpr);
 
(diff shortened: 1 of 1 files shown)


ast/src/main/java/com/graphicsfuzz/common/typing/Typer.java
```diff
@@ -102,7 +103,59 @@ public void visitFunctionDefinition(FunctionDefinition functionDefinition) {
   public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
     super.visitFunctionCallExpr(functionCallExpr);
 
-    // First, see if there is a builtin with a matching prototype.
+    // First, see whether this is an invocation of a GraphicsFuzz macro.  If it does, we handle it
```
paulthomson: If it does -> If it is ?

afd: Good catch.

Issue #842, opened 2020-01-13
When upgrading shader version, only rename conflicting symbols


jarisiru:
In order to avoid collisions with builtins, the shader upgrader renames all user-defined variables and functions. This could be made more elegant by only checking collisions with builtins of the target shader version.

PR #844, opened 2020-01-13
Shader upgrading: rename variables and function names

jarisiru:
Also moved the util from ast tree to common tree, cleaned up some placeholder bits, added command line option to skip renaming.


common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -46,14 +49,23 @@
...
+import java.util.Map;
 import net.sourceforge.argparse4j.ArgumentParsers;
+import net.sourceforge.argparse4j.impl.Arguments;
 import net.sourceforge.argparse4j.inf.ArgumentParser;
 import net.sourceforge.argparse4j.inf.ArgumentParserException;
 import net.sourceforge.argparse4j.inf.Namespace;
 
 public class UpgradeShadingLanguageVersion extends ScopeTrackingVisitor {
 
+  // flag for disabling renames
```
afd: 'flag' -> 'Flag'.

But it would be better to pass this as a command-line argument, by adding a --no-rename optional argument.

afd: Oh, I get it now - you're using this as a flag for bitmask stuff.  Can you update the comment to say something to that effect?

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -74,37 +86,84 @@
    */
   private List<Stmt> globalVariableInitializers = new ArrayList<>();
 
+  // Variable name renaming map
```
afd: I would change the comment to say that it maps a variable from its original name to a new name (and similar for functionRename).

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -74,37 +86,84 @@
...
+        .help("Do not rename user-defined variables and functions")
+        .action(Arguments.storeTrue());
+
     return parser.parseArgs(args);
   }
 
   public static void main(String[] args) throws IOException, InterruptedException {
 
     try {
       Namespace ns = parse(args);
```
afd: I would comment that you are using 'flags' as a bit mask.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -200,13 +270,22 @@ public void visitVariablesDeclaration(VariablesDeclaration variablesDeclaration)
...
+        variableRename.put(vdi.getName(), newname);
+        vdi.setName(newname);
+      }
       if (vdi.hasArrayInfo()) {
         visit(vdi.getArrayInfo().getSizeExpr());
       } else if (baseType instanceof ArrayType) {
         visit(((ArrayType) baseType).getArrayInfo().getSizeExpr());
       }
       if (vdi.hasInitializer()) {
         visit(vdi.getInitializer());
```
afd: Please format this as:

// TODO(https://github.com/google/graphicsfuzz/issues/842) Only rename if ...

i.e., put the URL in the parentheses.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -220,53 +299,61 @@ public void visitVariablesDeclaration(VariablesDeclaration variablesDeclaration)
...
+      functionCallExpr.setCallee(functionRename.get(functionCallExpr.getCallee()));
     }
     for (Expr e : functionCallExpr.getArgs()) {
       visit(e);
     }
   }
 
+  @Override
+  public void visitFunctionPrototype(FunctionPrototype functionPrototype) {
+
```
afd: Similar comment here for the formatting of TODO.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -220,53 +299,61 @@ public void visitVariablesDeclaration(VariablesDeclaration variablesDeclaration)
...
+    for (ParameterDecl p : functionPrototype.getParameters()) {
+      visit(p);
+    }
+  }
+
+  @Override
+  public void visitParameterDecl(ParameterDecl parameterDecl) {
+    visit(parameterDecl.getType());
+    if (parameterDecl.getName() != null) {
+      if ((flags & NORENAME) == 0) {
```
afd: And here.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done. The reason I'm using a bit mask instead of touching the parameter parser everywhere is that when using the util from code, you don't (probably) want to drag the parameter parser along.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done. Also, this comment was in a wrong place, moved it to the right one.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done.

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -74,37 +86,87 @@
...
+  private Map<String, String> variableRename = new HashMap<>();
+
+  // Function name renaming map. Maps original names with new names.
+  private Map<String, String> functionRename = new HashMap<>();
+
+  /*
+   * Originally texture fetch functions included information about the kind of sampler
+   * to be used, but later on these were simplified and the remaining texture fetch
+   * functions work with various samplers.
+   */
```
afd: I am not keen on the use of a 2D array for what is intuitively a String-to-String map.

I think it would be better to use a Map<String, String> here, and either initialize it in the constructor via a series of "put" operations, or do:

final Map<String, String> textureFunctionRenames = new HashMap<String, String>() {{
        put("shadow1D", "texture");
        put("shadow2D", "texture");
        ...
    }};

Alternatively, define a method to do the renaming:

private String renameTextureFunction(String originalName) {
  switch (originalName) {
    case "shadow1D":
    case "shadow2D":
         ...
    case "textureCube":
      return "texture";
    // similar for other cases
}

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
```diff
@@ -46,14 +49,23 @@
...
 import net.sourceforge.argparse4j.inf.Namespace;
 
 public class UpgradeShadingLanguageVersion extends ScopeTrackingVisitor {
 
+  // Flag for disabling renames. Used in a bit mask for conversion options.
+  static final int NORENAME = 1;
+
+  // Conversion flags. Used as a bit mask.
+  private int flags;
+
```
paulthomson: Have a boolean field `bool renameUserDefined`. Add the same to the constructor.

jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/UpgradeShadingLanguageVersion.java
jarisiru: Done as a bunch of put calls in a separate function.

PR #860, opened 2020-01-28
gfauto: improve download/run_cts...

paulthomson:
Improve `run_cts_gf_tests` and `download_cts_gf_tests`. Now runs spirv-opt with -O, -Os, and a custom, longer set of arguments.

Refactored to allow extracting shaders from .amber files without having to download the .amber files from CTS. Useful if you want to manually add some .amber files that are not yet submitted to the CTS.

### DIFF
--- a/gfauto/gfauto/download_cts_gf_tests.py
+++ b/gfauto/gfauto/download_cts_gf_tests.py
@@ -35,8 +35,8 @@ from gfauto import (
 
 
 def download_cts_graphicsfuzz_tests(  # pylint: disable=too-many-locals;
-    git_tool: Path, cookie: str, binaries: binaries_util.BinaryManager
-) -> None:
+    git_tool: Path, cookie: str, output_shaders_dir: Path,
+) -> Path:
     work_dir = Path() / "temp" / ("cts_" + fuzz.get_random_name())
 
     latest_change = gerrit_util.get_latest_deqp_change(cookie)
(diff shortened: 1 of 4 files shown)


gfauto/gfauto/download_cts_gf_tests.py
```diff
@@ -35,8 +35,8 @@
```
afd: Worth a comment on what this function returns?

gfauto/gfauto/download_cts_gf_tests.py
```diff
@@ -99,17 +99,26 @@ def download_cts_graphicsfuzz_tests(  # pylint: disable=too-many-locals;
...
         / "vulkancts"
         / "data"
         / "vulkan"
         / "amber"
         / "graphicsfuzz",
-        Path() / "graphicsfuzz",
+        output_shaders_dir,
     )
 
+
```
afd: Am I right that this is something only available to Khronos group members?  Worth documenting that explicitly?

afd: (I see that this was already present and just moved in this PR.)

gfauto/gfauto/download_cts_gf_tests.py
paulthomson: I've been using a slightly weird convention where most functions typically return the "output_path" parameter. This allows you to do things like:

```python
foo_dir = f(a, b, c, output_dir=work / "foo")
bar_dir = g(foo_dir)
```

But still, there a lot of missing return comments!

paulthomson: I added a coding conventions file. There are still a lot of missing docstrings, but I don't think I will fix this one in this PR.

gfauto/gfauto/download_cts_gf_tests.py
paulthomson: Done

PR #870, opened 2020-02-12
Implement loop truncation transformation

jarisiru:


common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
```diff
@@ -39,9 +64,90 @@
...
-  }
+    for (TranslationUnit tu : shaderJob.getShaders()) {
+      Declaration firstNonPrecisionDeclaration = null;
+      for (Declaration decl : tu.getTopLevelDeclarations()) {
+        if (decl instanceof PrecisionDeclaration) {
+          continue;
+        }
+        firstNonPrecisionDeclaration = decl;
+        break;
+      }
```
afd: Assert that firstNonPrecisionDeclaration is not null at this point, as we should always find something.

common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
```diff
@@ -39,9 +64,90 @@
...
+      // Add loop bound variable
+      tu.addDeclarationBefore(new VariablesDeclaration(new QualifiedType(BasicType.INT,
+              Arrays.asList(TypeQualifier.CONST)),new VariableDeclInfo(loopBoundName,null,
+              new Initializer(new IntConstantExpr(new Integer(loopLimit).toString())))),
+          firstNonPrecisionDeclaration);
```
afd: Google style for Java requires whitespace after commas; can you add please?  (I would have thought CheckStyle would flag this up but apparently it does not.)

afd: (And please do this throughout.)

common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
```diff
@@ -39,9 +64,90 @@
...
+            forStmt.setCondition(buildCondition(forStmt.getCondition()));
+            // Add block statement if it's missing
+            if (!(forStmt.getBody() instanceof BlockStmt)) {
+              forStmt.setBody(new BlockStmt(Arrays.asList(forStmt.getBody()), true));
+            }
+            // Add loop count increment a start of body block
+            ((BlockStmt) forStmt.getBody()).insertStmt(0,
+                new ExprStmt(new UnaryExpr(new VariableIdentifierExpr(loopCountName),
+                    UnOp.POST_INC)));
+          }
```
afd: Can you introduce:

private void handleLoopStmt(LoopStmt loopStmt) {
  // Do the condition check, replacement, and new statement adding
}

Because the same work is required for all three kinds of loop (including do..while, I think, as per my comment below), you could then do the "for" case as folows:

public void visitForStmt(ForStmt forStmt) {
  handleLoopStmt(forStmt);
  super.visitForStmt(forStmt);
}

and similar for the other cases.

common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
jarisiru: Went through the whole code and found only the one instance. And yes, I would have expected CheckStyle to have caught that.

common/src/main/java/com/graphicsfuzz/common/util/GloballyTruncateLoops.java
jarisiru: All three loop types were handled in a slightly different way - do-while always has body with a block statement, but it won't hurt to check for it. while... has a diffent scope rule, but that can be handled via a simple parameter, which I added.

PR #876, opened 2020-02-20
Integrate variable initializers in reducer

afd:
This gives the reducer the option, when semantics are not being
preserved, to initialize every variable in the shader during
reduction, to try to avoid uninitialized reads in reduced shaders.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
@@ -18,6 +18,7 @@ package com.graphicsfuzz.reducer;
 
 import com.graphicsfuzz.common.transformreduce.GlslShaderJob;
 import com.graphicsfuzz.common.transformreduce.ShaderJob;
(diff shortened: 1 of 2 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
```diff
@@ -289,15 +309,21 @@ private boolean isInteresting(ShaderJob state,
...
     String hash = null;
     if (useCache) {
+      // The cache is enabled, so first check for a cache hit.
+
+      // Write the state out to a shader job file, without doing any post-processing
+      // transformations.  This is because two different shader jobs might get post-processed to the
+      // same thing, and we want to avoid treating this as a reduction loop (and we *do* want to
+      // guard against reduction loops).
+      writeState(state, shaderJobFile, false, false, false, false);
+
```
paulthomson: Nice catch!

afd: :)

PR #877, opened 2020-02-20
Simplify literals more aggressively in the reducer

afd:
Before this change, the reducer would not attempt to simplify
literals, so that e.g. 8092.6 would remain in a reduced shader, even
if in dead code. This was based on the assumption that the exact form
of final literals would not really be important. However, having
worked a lot with the reducer it is certainly the case that a final
shader that does not involve weird literal values is easier to
understand than one that does. This change allows the reducer to
simplify any literal that is not in canonical form to one that is.

We should keep an eye on the knock-on effect this has on reducer
performance; by creating more reduction opportunities we could see a
performance drop.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/SimplifyExprReductionOpportunities.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/SimplifyExprReductionOpportunities.java
@@ -19,6 +19,7 @@ package com.graphicsfuzz.reducer.reductionopportunities;
 import com.graphicsfuzz.common.ast.IAstNode;
 import com.graphicsfuzz.common.ast.TranslationUnit;
 import com.graphicsfuzz.common.ast.decl.FunctionPrototype;
+import com.graphicsfuzz.common.ast.decl.VariableDeclInfo;
 import com.graphicsfuzz.common.ast.expr.Expr;
 import com.graphicsfuzz.common.ast.expr.FunctionCallExpr;
 import com.graphicsfuzz.common.ast.expr.VariableIdentifierExpr;
@@ -36,6 +37,10 @@ abstract class SimplifyExprReductionOpportunities
   final Typer typer;
 
   private boolean inLiveInjectedStmtOrDeclaration;
+
+  // Tracks whether we are visiting the components of a loop limiter's declaration (Which includes
+  // its initializer).  We do not wish to mess with these.
+  private boolean inLoopLimiterVariableDeclInfo;
 
   SimplifyExprReductionOpportunities(
         TranslationUnit tu,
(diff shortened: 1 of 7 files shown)


afd:
@paulthomson I resurrected this from a previous PR now that the reducer guards against out-of-bounds accesses during a semantics-changing reduction.

There is still the possibility of an out-of-bounds array access being introduced if:

live-injected code indexes into an array of size 1
the expression used for indexing is simplified to the literal 1
This could be solved by changing canonical constants from 1 to 0; I am reluctant to do that due to it introducing a much higher rate of division-by-zero.

I think the likelihood of this being a problem is really small (as we don't use any shaders with arrays of size 1), so I'm inclined to leave it as is and see whether it bites.

Regarding making the better simplification of literals a clean-up pass: that would be a bit complicated, and my hunch is that having it be a regular pass will not add too much overhead, since currently the reducer already tries turning all non-constant expressions into constants. I suggest we leave it as part of the core "ExprToConstant" pass and keep an eye on reduction times.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/SimplifyExprReductionOpportunities.java
```diff
@@ -37,6 +38,10 @@
 
   private boolean inLiveInjectedStmtOrDeclaration;
 
+  // Tracks whether we are visiting the components of a loop limiter's declaration (Which includes
```
paulthomson: Which -> which

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/ExprToConstantReductionOpportunitiesTest.java
```diff
@@ -98,4 +98,83 @@ public void testSwitch() throws Exception {
...
+  public void testSimplifyLiterals() throws Exception {
+    final String program = "#version 310 es\n"
+        + "struct S {\n"
+        + "  int a;\n"
+        + "  float b;\n"
+        + "};\n"
+        + "void main() {\n"
+        + "  100;\n"
+        + "  200.0;\n"
+        + "  300u;\n"
```
paulthomson: What if you have a switch with a few different int cases? Will they all get simplified to the same thing?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/SimplifyExprReductionOpportunities.java
afd: Fixed, thanks.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/ExprToConstantReductionOpportunitiesTest.java
afd: This was already catered for in ReductionOpportunitiesBase, via:

```
  @Override
  public void visitExprCaseLabel(ExprCaseLabel exprCaseLabel) {
    // Do *not* invoke super.visitExprCaseLabel(...), as we do not wish to look for opportunities
    // to simplify the literal expression that is used as a case label.  Literals are simple enough
    // already, and attempting to change the value of a case label literal runs the risk of
    // introducing duplicate case labels.
  }
```

However, I have added an extra test for it just to be sure.

PR #878, opened 2020-02-21
Reducer: be more aggressive in eliminating return statements

afd:
This change allows the reducer to eliminate non-void return statements
that are inside dead code blocks.


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/Compatibility.java
```diff
@@ -91,6 +91,18 @@ public static boolean compatible(Class<? extends IReductionOpportunity> first,
       return false;
     }
 
+    if (matchesEitherDirection(first, second,
+        StmtReductionOpportunity.class::equals,
+        UnswitchifyReductionOpportunity.class::equals)) {
+      return false;
+    }
```
paulthomson: This looks identical to the above. Can remove? Maybe could also add some comments to the ones where you can remember the reasons?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/InjectionTrackingVisitor.java
```diff
@@ -0,0 +1,149 @@
...
+import com.graphicsfuzz.common.ast.stmt.BreakStmt;
+import com.graphicsfuzz.common.ast.stmt.ExprCaseLabel;
+import com.graphicsfuzz.common.ast.stmt.IfStmt;
+import com.graphicsfuzz.common.ast.stmt.Stmt;
+import com.graphicsfuzz.common.ast.stmt.SwitchStmt;
+import com.graphicsfuzz.common.typing.ScopeTrackingVisitor;
+import com.graphicsfuzz.common.util.MacroNames;
+import com.graphicsfuzz.util.Constants;
+
+public class InjectionTrackingVisitor extends ScopeTrackingVisitor {
```
paulthomson: New class. Could add some documentation. Does this track when we are in injected code?

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/Compatibility.java
afd: Removed the duplication - I guess that must have arisen due to a previous change introducing the same compatibility check.

Unfortunately the reasons for these various checks have escaped me, and it wouldn't surprise me if some of them are no longer needed, due to having changed how the reducer works.  It would be worth digging into them in depth again at some point, removing those that do not seem to be needed and documenting why the others are needed.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/InjectionTrackingVisitor.java
afd: Done - the class was actually just extracted from ScopeTrackingVisitor (so not new code) - but I added some comments to bits of it, and quite a lot of commentary to InjectionTracker.

PR #879, opened 2020-02-21
Get array constructor support right in shading languages

afd:
Fixes an issue where shading language support for array constructors
was incorrect, which was leading to array constructors not being
generated.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
@@ -151,7 +151,11 @@ public interface ShadingLanguageVersion {
 
   boolean isWebGl();
 
-  boolean restrictedArrayIndexing();
+  /**
+   * GLSL versions 1.2+ and ESSL versions 3.0+ support array constructors.
+   * @return true if and only if array constructors are supported.
+   */
+  boolean supportedArrayConstructors();
 
   boolean restrictedForLoops();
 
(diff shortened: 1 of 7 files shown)


ast/src/main/java/com/graphicsfuzz/common/glslversion/ShadingLanguageVersion.java
```diff
@@ -151,7 +151,11 @@ static ShadingLanguageVersion webGlFromVersionString(String versionString) {
 
   boolean isWebGl();
 
-  boolean restrictedArrayIndexing();
+  /**
+   * GLSL versions 1.2+ and ESSL versions 3.0+ support array constructors.
+   * @return true if and only if array constructors are supported.
+   */
+  boolean supportedArrayConstructors();
```
paulthomson: Should we add some very simple tests?

paulthomson: I suppose the new shaders act as tests

afd: Hmm, I suppose we could have some tests that go through the claimed properties supported by shading languages and generate a simple shader that has the property, checking that it is valid if and only if we deem the property supported.

I'll open an issue for that, rather than it being something we do in this PR (as it is relevant to all the other properties too.)

afd: Opened #882 for this - and will merge this PR now.

PR #891, opened 2020-02-27
Only use donors of the same shader kind as the reference

afd:
Before this change, only .frag shaders were being used as donors, and
they were being used as donors for .vert and .comp shaders.

This change forces the generator to consider only shaders of the same
kind as the reference for donation into the reference.

Because this allows .comp shaders to be donated into .comp shaders,
which was not happening before, it exposed some limitations relating
to compute shader donation, specifically that the various compute
shader builtin variables were not being handled correctly, and that
interface blocks were not being handled in donation.  A similar issue,
with builtin variables, applied to vertex shaders.

This change introduces solutions to those problems.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateLiveCodeTransformation.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateLiveCodeTransformation.java
@@ -45,6 +45,7 @@ import com.graphicsfuzz.generator.util.RemoveReturnStatements;
 import com.graphicsfuzz.generator.util.TransformationProbabilities;
 import com.graphicsfuzz.util.Constants;
 import java.io.File;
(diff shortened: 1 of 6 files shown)


generator/src/main/java/com/graphicsfuzz/generator/transformation/DonateLiveCodeTransformation.java
```diff
@@ -45,6 +45,7 @@
 import com.graphicsfuzz.generator.util.TransformationProbabilities;
 import com.graphicsfuzz.util.Constants;
 import java.io.File;
+import java.lang.reflect.Array;
```
paulthomson: Remove this?

afd: Good spot!

PR #895, opened 2020-02-28
Account for void parameters

afd:
Some shaders in the wild use (void) to indicate no arguments to a
function; this change handles this by turning such declarations into
zero-argument functions when building the AST.

### DIFF
--- a/ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
+++ b/ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
@@ -497,4 +497,30 @@ public class PrettyPrinterVisitorTest {
         + "}").contains(ParseHelper.END_OF_GRAPHICSFUZZ_DEFINES));
   }
 
+  @Test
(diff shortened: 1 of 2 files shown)


ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
```diff
@@ -497,4 +497,30 @@ public void testGraphicsFuzzMacrosDueToSwitchMacro() throws Exception {
...
+    // that intent.
+
+    final String program = ""
+        + "int foo(void)\n"
+        + "{\n"
+        + " return 2;\n"
+        + "}\n"
+        + "void main(void)\n"
+        + "{\n"
+        + "}\n";
```
paulthomson: What about calls using void as well? E.g. `foo(void);`

afd: Turns out this is already captured by the way the grammar is structured, but I have added a test to demonstrate that.

PR #899, opened 2020-03-06
gfauto: support stable shaders

paulthomson:
With this change, only reference shader jobs that start with "stable_" will lead to wrong image tests. Other tests will just render the variant shader to see if it crashes. This relies on there being shader jobs that are prefixed with "stable_". This change does not bias the random choice of reference towards "stable" shaders.


gfauto/gfauto/fuzz_glsl_test.py
```diff
@@ -179,18 +195,34 @@ def make_test(
...
+    test = Test(glsl=TestGlsl(spirv_opt_args=spirv_opt_args), derived_from=derived_from)
 
     test.binaries.extend([binary_manager.get_binary_by_name(name="glslangValidator")])
     add_spirv_shader_test_binaries(test, spirv_opt_args, binary_manager)
 
     # Write the test metadata.
     test_util.metadata_write(test, subtest_dir)
 
+    # If the reference shader is "stable" (with respect to floating-point sensitivity)
+    # then this can be a wrong image test. I.e. we will render the reference and variant shaders
```
afd: Nit: I would not start a sentence "I.e." - I would change "test. I.e." to "test; i.e." here.

gfauto/gfauto/fuzz_glsl_test.py
```diff
@@ -179,18 +195,34 @@ def make_test(
     subtest_dir: Path,
     spirv_opt_args: Optional[List[str]],
     binary_manager: binaries_util.BinaryManager,
+    derived_from: Optional[str],
```
afd: Maybe worth a comment (here or elsewhere) on what 'derived_from' means.

gfauto/gfauto/fuzz_spirv_test.py
```diff
@@ -52,17 +52,36 @@ def make_test(
...
+    )
 
     fuzz_glsl_test.add_spirv_shader_test_binaries(test, spirv_opt_args, binary_manager)
 
     # Write the test metadata.
     test_util.metadata_write(test, subtest_dir)
 
+    # If the reference shader is "stable" (with respect to floating-point sensitivity)
+    # then this can be a wrong image test. I.e. we will render the reference and variant shaders
+    # and compare the output images.
```
afd: Same comment about I.e.

gfauto/gfauto/fuzz_glsl_test.py
afd: OK, I see you do comment it in the protobuf definition.

PR #902, opened 2020-03-08
Fixed bug in vectorizer related to single statement for loops.

afd:
Fixes a problem where the tracking of scopes by the vectorizer was not working properly for single-statement for (and while) loops, which introduce new scopes even though they do not have blocks as bodies.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/typing/ScopeTrackingVisitor.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/typing/ScopeTrackingVisitor.java
@@ -90,7 +90,7 @@ public abstract class ScopeTrackingVisitor extends StandardVisitor {
     //
     // because it declares 'i' twice in the same scope.
     //
-    // Furthermore, a 'while' statement introduces a new scope regardless of whether its body is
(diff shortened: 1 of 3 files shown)


ast/src/main/java/com/graphicsfuzz/common/typing/ScopeTrackingVisitor.java
```diff
@@ -90,7 +90,11 @@ public void visitForStmt(ForStmt forStmt) {
...
     //
-    // We thus have to push a new scope before traversing a 'for' statement and pop it afterwards.
+    // Furthermore, a 'while' statement introduces a new scope regardless of whether its body is
+    // a block or a single statement.
+    //
+    // We thus have to push a new scope before traversing a 'for' statement and pop it afterwards,
+    // and we do this regardless of whether the body is a block.
     pushScope();
     super.visitForStmt(forStmt);
     popScope();
```
paulthomson: Did you mean to talk about "while" here?

afd: No - good catch, thanks!

PR #904, opened 2020-03-08
Rework code donation

afd:
This change reworks code donation and makes it dramatically simpler.  Various mechanisms were in place to try to allow the sharing of struct types between the reference and donor, and to regard shaders as incompatible if they had identically-named but structurally different structs.  This change forces all structs to be renamed when processing a donor, so that shaders can no longer be incompatible for this reason.  It means that structs are never shared between reference and donor, which may in some cases lead to less interesting shaders, but it makes the donation process significantly simpler.


afd:
This one turns out to depend on some of the others (I made it from a cherry-picked commit) - will update it once they are merged.

PR #925, opened 2020-03-13
Update Fuzzer API

### LINKED ISSUE #922: Vulkan fuzzer layer uses outdated fuzzer API
If you try to compile [vulkan layers](https://github.com/google/graphicsfuzz/tree/master/third_party/vulkan_layers) using KhronosGroup/SPIRV-Tools@1fe9bcc1 commit you will get a compilation error. The reason, as it seems, is that `Fuzzer`'s API has been changed. Consider:

the API 
https://github.com/KhronosGroup/SPIRV-Tools/blob/1fe9bcc10824c1fa35bd9b697188340132d39213/source/fuzz/fuzzer.h#L44

it's usage https://github.com/google/graphicsfuzz/blob/ed7af5e223101f3117c0498cb2752e853bda6290/third_party/vulkan_layers/shader_fuzzer/layer_impl.cpp#L116

Vasniktel:
Fixes #922.
Note that I experienced compilation issues with `protobuf` library.
It looks like `protobuf` adds debug postfix to its target so that we are getting `libprotobufd.a` instead of `libprotobuf.a`. This causes build script to crash saying that the latter target is undefined. To fix use:
```
$ mkdir build && cd build
$ cmake -Dprotobuf_DEBUG_POSTFIX="" ..
```

### DIFF
--- a/third_party/vulkan_layers/.gitignore
+++ b/third_party/vulkan_layers/.gitignore
@@ -1,2 +1,4 @@
 cmake-build-debug
 .idea
+
+build/
--- a/third_party/vulkan_layers/shader_fuzzer/layer_impl.cpp
+++ b/third_party/vulkan_layers/shader_fuzzer/layer_impl.cpp
@@ -113,17 +113,15 @@ TryFuzzingShader(VkShaderModuleCreateInfo const *pCreateInfo) {
   }
 
   // Create a fuzzer and the various parameters required for fuzzing.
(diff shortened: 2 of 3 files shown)


Vasniktel:
@googlebot I signed it!

Vasniktel:
Ok, thanks for letting me know. Changes will probably be required anyway, though, since this PR is probably obsolete.

Vasniktel:
This functionality has already been implemented in #1045.

PR #926, opened 2020-03-17
Change vulkan layer common module library type

ilkkasaa:
Changes the vulkan layers' common module's library type from STATIC to SHARED.
Using static causes some linker error. This is just a workaround.

Using static library causes following linker error when building amber-scoop and shader-fuzzer vulkan layer:
```
/usr/bin/ld: ../common/libvulkan_layers_common.a(spirv_util.cpp.o): relocation R_X86_64_PC32 against symbol `_ZNSt4pairIb14spv_target_envEC1IbRS0_Lb1EEEOT_OT0_' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
```

### DIFF
--- a/third_party/vulkan_layers/common/CMakeLists.txt
+++ b/third_party/vulkan_layers/common/CMakeLists.txt
@@ -19,7 +19,7 @@ set(VULKAN_LAYERS_COMMON_SOURCES
         spirv_util.cpp
         )
 
-add_library(vulkan_layers_common STATIC ${VULKAN_LAYERS_COMMON_SOURCES})
+add_library(vulkan_layers_common SHARED ${VULKAN_LAYERS_COMMON_SOURCES})
 
 target_include_directories(vulkan_layers_common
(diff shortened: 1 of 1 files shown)


third_party/vulkan_layers/amber_scoop/layer_impl.cpp
```diff
@@ -3,7 +3,7 @@
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * You may obtain a copy of the Licens
```
paulthomson: This should be fixed :)

ilkkasaa: Oops :) I didn't even mean to add the second commit to the PR.

afd:
Closing this as we are moving the layers over to gf-layers.

PR #933, opened 2020-03-23
Be more liberal with uniform pruning

afd:
Instead of throwing an exception if a reference shader has too many
uniforms, or if it is not possible to bring the number of uniforms in
a variant shader down to size by pruning injected uniforms, this
change makes uniform pruning more liberal so that shaders are always
pruned down to size by pruning original uniforms if needed.

### DIFF
--- a/generator/src/main/java/com/graphicsfuzz/generator/tool/PrepareReference.java
+++ b/generator/src/main/java/com/graphicsfuzz/generator/tool/PrepareReference.java
@@ -138,9 +138,10 @@ public final class PrepareReference {
     }
 
     if (maxUniforms > 0) {
-      // The reference shader job has more uniforms than the specified limit, so prune some of them
-      // by inlining their values into the shaders.
-      PruneUniforms.prune(shaderJob, maxUniforms, Collections.emptyList());
+      // If the reference shader job has more uniforms than the specified limit, some of them must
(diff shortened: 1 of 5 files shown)


generator/src/main/java/com/graphicsfuzz/generator/tool/PrepareReference.java
```diff
@@ -135,8 +137,10 @@ public static void prepareReference(ShaderJob shaderJob,
           shaderJob.getPipelineInfo());
     }
 
-    if (maxUniforms > 0 && shaderJob.getPipelineInfo().getNumUniforms() > maxUniforms) {
-      throw new RuntimeException("Too many uniforms in reference shader job.");
+    if (maxUniforms > 0) {
+      // The reference shader job has more uniforms than the specified limit, so prune some of them
+      // by inlining their values into the shaders.
+      PruneUniforms.prune(shaderJob, maxUniforms, Collections.emptyList());
```
paulthomson: This comment is a bit confusing, as it is not necessarily true. And perhaps `prune` should be `pruneIfNeeded`?

afd: Agreed on both counts - done.

PR #938, opened 2020-03-27
One variable per line and matrix support

jarisiru:
Split in/out variables to separate lines if more than one
is declared on the same line, and also added support for
matrix variables.


common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderToShaderJob.java
```diff
@@ -117,49 +117,71 @@ public static ShaderJob createShaderJob(TranslationUnit tu, IRandom generator) {
...
+          // Increment location index
+          if (basicType == BasicType.MAT2X2
+              || basicType == BasicType.MAT2X3
+              || basicType == BasicType.MAT2X4
+              || basicType == BasicType.MAT3X2
+              || basicType == BasicType.MAT3X3
+              || basicType == BasicType.MAT3X4
+              || basicType == BasicType.MAT4X2
+              || basicType == BasicType.MAT4X3
+              || basicType == BasicType.MAT4X4) {
```
afd: For this can you instead do:

```
if (BasicTypes.allMatrixtypes().contains(basicType)) { ... }
```

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderToShaderJob.java
```diff
@@ -183,6 +205,105 @@ public static ShaderJob createShaderJob(TranslationUnit tu, IRandom generator) {
                 new FloatConstantExpr(Float.toString(generator.nextFloat())),
                 new FloatConstantExpr(Float.toString(generator.nextFloat())),
                 new FloatConstantExpr(Float.toString(generator.nextFloat()))));
+          } else if (basicType == BasicType.MAT2X2) {
```
afd: I suggest you simplify all this code in the following manner:

```
final List<Expr> constructorParams = new ArrayList<>();
if (basicType.getElementType() == BasicType.INT) {
  for (int i = 0; i < basicType.getNumElements(); i++) {
     // add an appropriate IntConstantExpr
  }
} else if (basicType.getElementType() == BasicType.UINT) {
   // similar, but for a uint constant expression - don't forget to put 'u' at the end of the literal
} else if {
   // similar for the FLOAT and BOOL cases
}
final Expr initExpr = new TypeConstructorExpr(basicType.toString(), constructorParams);
```

That would handle the boolean/bvec case, as well as doing all the other cases more compactly.

PR #948, opened 2020-04-06
Smarter removal of loops with limiters

afd:
The statement removal reduction pass is enhanced with an analysis that
figures out those loops whose removal will not affect the limiting
(via loop limiters) of other loops.  This is used when looking for
statements to remove: a live-injected loop whose removal is known not
to impact the limiting of other loops can be removed.

### DIFF
--- a/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/StmtReductionOpportunities.java
+++ b/reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/StmtReductionOpportunities.java
@@ -439,7 +439,7 @@ public class StmtReductionOpportunities
       // outside any loop).
       for (int i = loopStack.size() - 1; i >= 0; i--) {
         if (loopStack.get(i).getRight().contains(variablesDeclaration)) {
-          // This is where the loop limiter is declared; shallow loops are not affected by this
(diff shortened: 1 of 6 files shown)


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/StmtReductionOpportunities.java
```diff
@@ -331,4 +380,109 @@ private boolean isEmptyAndUnreferencedDeclaration(DeclarationStmt stmt) {
...
+              .getVariablesDeclaration();
+
+      // Walk the loop stack backwards until we find the loop in which this limiter was declared
+      // (or until we reach the bottom of the stack, in the case that the limiter is declared
+      // outside any loop).
+      for (int i = loopStack.size() - 1; i >= 0; i--) {
+        if (loopStack.get(i).getRight().contains(variablesDeclaration)) {
+          // This is where the loop limiter is declared; shallow loops are not affected by this
+          // loop limiter.
+          break;
```
paulthomson, line 444: Do you mean "shallower" loops?

afd: Yes, good catch.

PR #956, opened 2020-04-09
Add $grid field handling to amber converter

jarisiru:


gfauto/gfauto/amber_converter.py
```diff
@@ -214,7 +214,7 @@ def amberscript_uniform_buffer_bind(uniform_json: str, prefix: str) -> str:
     result = ""
     uniforms = json.loads(uniform_json)
     for name, entry in uniforms.items():
-        if name == "$compute":
+        if name[0] == "$":
```
paulthomson, line 217: I prefer `name.startswith("$")`. Also, please update the doc comment above (since you did it already for the function below).

gfauto/gfauto/amber_converter.py
```diff
@@ -296,7 +296,7 @@ def amberscript_uniform_buffer_def(uniform_json_contents: str, prefix: str) -> s
 
     for name, entry in uniforms.items():
 
-        if name == "$compute":
+        if name[0] == "$":
```
paulthomson, line 299: `name.startswith("$")`

gfauto/gfauto/amber_converter.py
```diff
@@ -445,6 +455,9 @@ def to_shader_job(self) -> ShaderJob:
             must_exist=True,
         )
 
+        # Figure out if we want to draw rectangle or grid
```
paulthomson, line 458: Figure out if we want to draw rectangle or grid
->
Figure out if we want to draw a rectangle or grid. (Add "a"; end in full stop.)

PR #963, opened 2020-04-16
Improved for loop formatting and modified a test for it

Vihanakangas:
This change has modified for loop formatting in a way that the loop gets formatted to:
```
for( int i = 0; i < 1; i ++ )
{
}
```
Instead of it being formatted to:
```
for(
     int i = 0;
     i < 1;
     i ++
    )
 {
 }
```


ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -94,6 +94,7 @@
   private boolean inFunctionDefinition = false;
   private final boolean emitGraphicsFuzzDefines;
   private final Optional<String> license;
+  private boolean insideForStatement = false;
```
afd, line 97: Please rename to `insideForStatementHeader` (in relation to a comment below), and on the line above have a brief comment to explain that this allows us to format declarations differently if they are part of the header of a for statement.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -460,23 +464,22 @@ public void visitWhileStmt(WhileStmt whileStmt) {
...
     if (forStmt.hasIncrement()) {
       visit(forStmt.getIncrement());
     }
-    out.append(newLine());
-    out.append(indent() + ")" + newLine());
+    out.append(" )" + newLine());
     increaseIndent();
     visit(forStmt.getBody());
     decreaseIndent();
+    insideForStatement = false;
```
paulthomson, line 482: I think you probably want this `insideForStatement = false;` to go higher up, otherwise declaration statements inside the for loop body won't get a new line.

```
for(;;;){
int i;int y;int z;
}
```

That is my guess anyway. 

The boolean would also get overwritten to false after a nested for loop, so I assume this would look OK:

```
for(;;;){
for(;;;){}
int i;
int y;
int z;
}
```

Maybe rename it `insideForInitStatement` and just set it to true during `visit(forStmt.getInit());`.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -159,7 +160,10 @@ public void visitPrecisionDeclaration(PrecisionDeclaration precisionDeclaration)
   public void visitDeclarationStmt(DeclarationStmt declarationStmt) {
     out.append(indent());
```
afd, line 161: Do not do this indent if we are in a for statement header.

This will avoid a space after the opening paren of "for(" in "for(int i"

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -460,23 +464,22 @@ public void visitWhileStmt(WhileStmt whileStmt) {
...
     }
-    out.append(";" + newLine());
-    out.append("    " + indent());
+    out.append("; ");
     if (forStmt.hasIncrement()) {
       visit(forStmt.getIncrement());
     }
-    out.append(newLine());
-    out.append(indent() + ")" + newLine());
+    out.append(" )" + newLine());
```
afd, line 478: At this point you should set `insideForStatementHeader` to false, because we're done visiting the header and are now going to visit the body.  We do not want to keep this flag set while we process the body of the loop.

afd, line 478: Please remove the space in " )", so that we don't have a space between the end of the for loop header and its closing ")".

ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
```diff
@@ -165,11 +165,7 @@ public void testParseAndPrintComputeShader() throws Exception {
         + "layout(local_size_x = 128, local_size_y = 1) in;\n"
         + "void main()\n"
         + "{\n"
-        + " for(\n"
-        + "     uint d = gl_WorkGroupSize.x / 2u;\n"
-        + "     d > 0u;\n"
-        + "     d >>= 1u\n"
-        + " )\n"
+        + " for( uint d = gl_WorkGroupSize.x / 2u; d > 0u; d >>= 1u )\n"
```
afd, line 168: Remove two spaces - after "(" and before ")", in relation to the above comments.

afd: Similar fixes needed to other tests.

afd: At the end of this file, can you please add the following test which guards against declarations in a loop being put on the same line?

```
  @Test
  public void testParseAndPrintLoopWithDeclarationsInside() throws Exception {
    final String program = ""
        + "void main()\n"
        + "{\n"
        + " for(int i = 0; i < 10; i ++)\n"
        + "  {\n"
        + "   int a;\n"
        + "   int b;\n"
        + "  }\n"
        + "}\n";
    assertEquals(program, PrettyPrinterVisitor.prettyPrintAsString(ParseHelper.parse(program
    )));
  }
```

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
Vihanakangas: Done.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
Vihanakangas: Done.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
Vihanakangas: Fixed.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
Vihanakangas: Done.

Vihanakangas: Done.

ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
Vihanakangas: Done.

PR #970, opened 2020-04-21
Add --shader-stage to fuzz specific shader stage

paulthomson:


generator/src/main/java/com/graphicsfuzz/generator/tool/Generate.java
```diff
@@ -166,15 +170,24 @@ public static StringBuilder generateVariant(ShaderJob shaderJob,
                                               GeneratorArguments args,
                                               IRandom random) {
     final StringBuilder result = new StringBuilder();
+    List<TranslationUnit> shaders = new ArrayList<>(shaderJob.getShaders());
```
afd, line 173: Make final (unless there's a reason not to).

PR #974, opened 2020-04-27
Improvements to frag2vert conversion

jarisiru:
- Shader versions taken from source instead of hardcoding
- gl_FragDepth handling
- discard changed to null operation
- Various fragment shader-only functions changed to parenthesis operations (i.e, dFdy(x) -> (x))
-  _GLF_FragCoord and _GLF_FragDepth only injected if gl_FragCoord and gl_FragDepth are used


common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -77,6 +81,27 @@ public static ShaderJob convertShaderJob(ShaderJob sj) throws Exception {
     TranslationUnit vertexShader = sj.getFragmentShader().get();
     vertexShader.setShaderKind(ShaderKind.VERTEX);
 
+    // Analyse the shader for used features (can't use anonymous class due to member access)
+    class Analysis extends StandardVisitor {
+      public boolean usesFragCoord = false;
+      public boolean usesFragDepth = false;
```
afd, line 87: Please make private and add getUsesFragCoord() and getUsesFragDepth() methods.  A bit annoying I know, but I try to enforce that fields are private throughout the project.

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -77,6 +81,27 @@ public static ShaderJob convertShaderJob(ShaderJob sj) throws Exception {
     TranslationUnit vertexShader = sj.getFragmentShader().get();
     vertexShader.setShaderKind(ShaderKind.VERTEX);
 
+    // Analyse the shader for used features (can't use anonymous class due to member access)
+    class Analysis extends StandardVisitor {
```
afd, line 85: Rename to e.g. FragmentBuiltinUsageAnalysis.

Also make the class private (I think that should work and if so it is preferred).

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -77,6 +81,27 @@ public static ShaderJob convertShaderJob(ShaderJob sj) throws Exception {
...
+        if (variableIdentifierExpr.getName().equals(OpenGlConstants.GL_FRAG_COORD)) {
+          usesFragCoord = true;
+        }
+        if (variableIdentifierExpr.getName().equals(OpenGlConstants.GL_FRAG_DEPTH)) {
+          usesFragDepth = true;
+        }
+      }
+    }
+    
+    final Analysis analysis = new Analysis();
```
afd, line 101: Rename variable in accordance with how you rename the class.

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -124,44 +149,55 @@ void main() {
 
     assert firstNonPrecisionDeclaration != null;
 
-    // Add a global _GLF_FragCoord that will replace gl_Fragcoord
-    vertexShader.addDeclarationBefore(
-        new VariablesDeclaration(
-            BasicType.VEC4, new VariableDeclInfo(Constants.GLF_FRAGCOORD, null, null)),
-        firstNonPrecisionDeclaration);
+    if (analysis.usesFragDepth) {
```
afd, line 152: Use getter method; same below.

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -124,44 +149,55 @@ void main() {
...
+                      new TypeConstructorExpr(BasicType.VEC4.toString(),
+                          new FloatConstantExpr("128.0"),
+                          new FloatConstantExpr("128.0"),
+                          new FloatConstantExpr("1.0"),
+                          new FloatConstantExpr("1.0")),
+                      BinOp.MUL),
+                  BinOp.ASSIGN)));
+    }
+
+    final IParentMap parentMap = IParentMap.createParentMap(vertexShader);
```
afd, line 200: Please add a comment to explain the overall purpose of the anonymous visitor that follows.

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
```diff
@@ -195,6 +235,39 @@ public void visitVariablesDeclaration(VariablesDeclaration variablesDeclaration)
...
+      @Override
+      public void visitDiscardStmt(DiscardStmt discardStmt) {
+        // Re
+        parentMap.getParent(discardStmt).replaceChild(discardStmt, new NullStmt());
+      }
+
+      // Replace fragment-only functions with parenthesis expression
+      @Override
+      public void visitFunctionCallExpr(FunctionCallExpr functionCallExpr) {
+        // For most of the functions, they take only one parameter and it's the same
```
afd, line 249: I think you should do super.visitFunctionCallExpr(functionCallExpr) here.

Then you don't need to use the loop below to visit each of the child arguments.

Can you add a test to make sure that a nested expression such as: dFdx(dFdy(...))) gets handled correctly?  I'm pretty sure it does, but doing child replacement with nesting can be tricky so would be good to be sure.  (BTW: I'm not sure that you're allowed to nest these two functions in GLES actually!)

might not get handled correctly.

The outer dFdx would get handled first, becoming dFdy(a, b).  Then the dFdY(a, b) would

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: renamed, but private was not allowed

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: done

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderJobToVertexShaderJob.java
jarisiru: super call added, and I already had a test like that (gl_FragDepth = dFdx(0.01 + dFdy(0.02 + dFdxFine(0.03 + dFdyFine(0.04)))); etc). It shouldn't matter whether the inner or outer is handled first, because all it does is remove the function call itself, not even the parentheses.

PR #983, opened 2020-05-06
gfauto: fix SSBO binding

paulthomson:
The AmberScript generated from a compute shader job always had a binding location of 0. Now, it uses the binding location from the JSON.

Also add some missing words from the whitelist.dic file.

### DIFF
--- a/gfauto/whitelist.dic
+++ b/gfauto/whitelist.dic
@@ -242,3 +242,5 @@ writelines
 colorgrid
 logd
 gf
+SIGSEGV
+SIGABRT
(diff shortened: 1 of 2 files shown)


gfauto/whitelist.dic
```diff
@@ -242,3 +242,5 @@ writelines
 colorgrid
 logd
 gf
+SIGSEGV
+SIGABRT
```
afd, line 246: Was it intentional to include these as part of the PR?  (Fine if so.)

paulthomson: Yeah, I accidentally pushed another change straight to master (which should now be impossible). This fix is needed before anything will pass!

PR #986, opened 2020-05-13
Pass exra SPIRV-Tools args to spirv-opt

afd:


paulthomson:
I think I'd like to do it a different way (by just adding the common_spirv_args to spirv_opt_args at a certain point).

afd:
Sounds good - I'll close this.

PR #989, opened 2020-05-19
Add skeleton for literal to uniform reduction

paulthomson:


reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
```diff
@@ -132,8 +158,23 @@ public ReductionDriver(ReducerContext context,
...
     }
-    this.passManager = new SystematicReductionPassManager(initialPasses, corePasses, cleanupPasses);
+    return new SystematicReductionPassManager(initialPasses, corePasses, cleanupPasses);
+  }
+
+  private static IReductionPassManager getLiteralsToUniformsPassManager(
+      ReducerContext context,
+      boolean verbose) {
+    final List<IReductionPass> initialPasses = new ArrayList<>();
+    final List<IReductionPass> corePasses = new ArrayList<>();
```
afd, line 168: I would get rid of 'initialPasses' and 'corePasses' and use Collections.emptyList() where you use them.  If you don't want to do that then how about initializing them to Collections.emptyList() here?

reducer/src/main/java/com/graphicsfuzz/reducer/ReductionDriver.java
```diff
@@ -132,8 +158,23 @@ public ReductionDriver(ReducerContext context,
...
+    final List<IReductionPass> cleanupPasses = new ArrayList<>();
```
afd, line 169: I would either initialize this to Arrays.asList(new SystematicReductionPass(...)), or just pass Arrays.asList(...) where you currently use cleanupPasses.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
```diff
@@ -0,0 +1,62 @@
...
+        .findOpportunities(shaderJob,
+            new ReducerContext(false, ShadingLanguageVersion.ESSL_100, new RandomWrapper(0),
+                new IdGenerator()));
+
+    assertEquals("There should be one opportunity", 1, ops.size());
+
+    ops.get(0).applyReduction();
+
+    //noinspection OptionalGetWithoutIsPresent
+    CompareAsts.assertEqualAsts(shaderReplaced, shaderJob.getFragmentShader().get());
```
afd, line 59: There should also be an assertion about the pipeline info to check that it has the new uniform in it.

PR #996, opened 2020-06-02
Literal to uniform

amakin-siru:
A feature that replaces integer literals with an element in a uniform array.


amakin-siru:
@googlebot I signed it!

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
 
   private final Expr literalExpr;
```
afd, line 31: It would be good to give this the more specific type, ConstantExpr, if possible.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
...
   private final PipelineInfo pipelineInfo;
+  private final TranslationUnit translationUnit;
+  private final int index;
```
afd, line 34: This field deserves a short comment to explain what it indexes into.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
...
+    this.translationUnit = translationUnit;
+  }
+
+  public Expr getLiteral() {
+    return this.literalExpr;
+  }
+
+  public int getIndex() {
+    return this.index;
   }
```
afd, line 55: I suggest removing these.  They shouldn't be needed.  If they are only used in tests then instead the test code should apply transformations and see that they have the right effect.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
...
 
   @Override
   void applyReductionImpl() {
-    // TODO: ...
+
+    IParentMap parentMap = IParentMap.createParentMap(translationUnit);
+
+    ArrayIndexExpr aie = new ArrayIndexExpr(new VariableIdentifierExpr("values"),
```
afd, line 62: The name "values" might already be defined.  Please add a new constant to Constants, with a suitable name (using the "_GLF_" prefix).

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
...
+
+  public int getIndex() {
+    return this.index;
   }
 
   @Override
   void applyReductionImpl() {
-    // TODO: ...
+
+    IParentMap parentMap = IParentMap.createParentMap(translationUnit);
```
afd, line 60: Here, and elsewhere - make local variables final unless it's not possible to do so.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
```diff
@@ -17,15 +17,78 @@
...
+import java.util.Optional;
 
 public final class LiteralToUniformReductionOpportunities {
 
   private LiteralToUniformReductionOpportunities() {
-    // This class just provides a static method.
+    // This class just provides a static methods.
+  }
+
+  static void replaceOpportunities(ShaderJob shaderJob) {
```
afd, line 46: This method should go, and the code to add the array and new pipeline info field should be in the "applyImpl" method of the reduction opportunity.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -25,18 +30,40 @@
...
+  }
+
+  public int getIndex() {
+    return this.index;
   }
 
   @Override
   void applyReductionImpl() {
-    // TODO: ...
+
```
afd, line 59: Here, you should check whether the required array already exists in the shader and pipeline info.  If it doesn't, you should bring the array into existence.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+    final String arrayName = Constants.INT_LITERAL_UNIFORM_VALUES;
+    final List<Number> values;
+    final int index;
+
+    // Creates a combined list of the current and previously added values.
+    // Updates the index number to point at the corresponding element in the array.
+    if (this.shaderJob.getPipelineInfo().hasUniform(arrayName)) {
```
afd, line 69: Except in a constructor or equals method, avoid unnecessary uses of "this".

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+      values = this.shaderJob.getPipelineInfo().getArgs(arrayName);
+      if (values.contains(((IntConstantExpr)this.literalExpr).getNumericValue())) {
```
afd, line 71: Leave one space after the closing ) of a cast.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
   @Override
   void applyReductionImpl() {
-    // TODO: ...
+
+    final String arrayName = Constants.INT_LITERAL_UNIFORM_VALUES;
+    final List<Number> values;
+    final int index;
+
+    // Creates a combined list of the current and previously added values.
+    // Updates the index number to point at the corresponding element in the array.
```
afd, line 68: I suggest you first check whether the array exists in the pipeline info and add an empty array if it does not.  Then your code for either finding an existing element or adding a new element will work for both cases.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+      arraySize.getArrayInfo().setConstantSizeExpr(2);
+      final InterfaceBlock ib =
+          new InterfaceBlock(Optional.of(new LayoutQualifierSequence(new SetLayoutQualifier(0),
+              new BindingLayoutQualifier(binding))),
+              TypeQualifier.UNIFORM,
+              "buf" + binding,
+              Collections.singletonList(arrayName),
+              Collections.singletonList(arraySize),
+              Optional.empty()
+          );
```
afd, line 121: Our AST framework currently expects uniforms to not be declared in interface blocks.  There is some special logic for turning them into interface blocks just before writing them out to file.  Thus you should instead just create a global with uniform qualifier.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+      // with increased size.
+      Optional<Declaration> oldDeclaration = Optional.empty();
+      final List<Declaration> declarations = tu.getTopLevelDeclarations();
+      for (Declaration declaration : declarations) {
+        if (declaration instanceof InterfaceBlock) {
+          final List<String> members = ((InterfaceBlock) declaration).getMemberNames();
+          if (members.get(0).equals(arrayName)) {
+            oldDeclaration = Optional.of(declaration);
+          }
+        }
```
afd, line 104: We should apply this transformation to shaders where uniforms are not in interface blocks - so just look out for a global uniform variable with the right name.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+
+      tu.addDeclaration(ib);
+
+      // Replaces the literal with an element in the uniform array.
+      final IParentMap parentMap = IParentMap.createParentMap(tu);
+      final ArrayIndexExpr aie = new ArrayIndexExpr(new VariableIdentifierExpr(arrayName),
+          new IntConstantExpr(String.valueOf(index)));
+
+      parentMap.getParent(this.literalExpr).replaceChild(this.literalExpr,
+          aie);
```
afd, line 131: A problem here is that the literal expression will not appear in *every* translation unit associated with the shader job, but only in one of them.

I therefore think that you should record the specific translation unit in which the literal appears as a field of this opportunity class, in addition to the shader job.  Then you should move this code to after the loop that goes through all the translation units, and have it do the parentmap thing to just the translation unit in which the literal expression exists.

I believe in its current form you would find you get an exception if you try this opportunity on a multi-shader shader job - it would be good to have a test for that.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
     extends AbstractReductionOpportunity {
 
-  private final Expr literalExpr;
-  private final PipelineInfo pipelineInfo;
+  private final ConstantExpr literalExpr;
+  private final TranslationUnit translationUnit;
+  private final ShaderJob shaderJob;
+  private final Number numericValue;
+  private final String arrayName;
+  private final BasicType basicType;
```
afd, line 49: These should not be fields.  They should not be required outside applyReductionImpl.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,120 @@
...
+      this.basicType = BasicType.FLOAT;
+      this.numericValue = Float.valueOf(((FloatConstantExpr) literalExpr).getValue());
+    }
+
+    if (!shaderJob.getPipelineInfo().hasUniform(this.arrayName)) {
+      shaderJob.getPipelineInfo().addUniform(this.arrayName, this.basicType,
+          Optional.of(0), new ArrayList<>());
+      shaderJob.getPipelineInfo().addUniformBinding(this.arrayName, false,
+          shaderJob.getPipelineInfo().getNumUniforms());
+    }
```
afd, line 78: This does not belong in the constructor.  We do not want to mutate the shader job when we create this opportunity - any mutations to the shader job should occur only if and when the opportunity is actually applied.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
afd: The code you have currently in the constructor should go here.

What are currently fields should be local variables.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
```diff
@@ -25,40 +26,152 @@
...
+        + "  int a = _GLF_uniform_int_values[1];"
+        + "}";
+
+    final List<TranslationUnit> shaders = new ArrayList<>();
+    shaders.add(ParseHelper.parse(vertexShader, ShaderKind.VERTEX));
+    shaders.add(ParseHelper.parse(fragmentShader, ShaderKind.FRAGMENT));
     final PipelineInfo pipelineInfo = new PipelineInfo();
     final ShaderJob shaderJob = new GlslShaderJob(Optional.empty(),
-        pipelineInfo, ParseHelper.parse(shader, ShaderKind.FRAGMENT));
+        pipelineInfo, shaders);
```
afd, line 66: Here, add:

```
assertEquals(0, pipelineInfo.getNumUniforms());
```

I believe this should pass at present.

However ... (see next comment)

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
```diff
@@ -25,40 +26,152 @@
...
 
-    List<LiteralToUniformReductionOpportunity> ops =
+    final List<LiteralToUniformReductionOpportunity> ops =
         LiteralToUniformReductionOpportunities
         .findOpportunities(shaderJob,
             new ReducerContext(false, ShadingLanguageVersion.ESSL_100, new RandomWrapper(0),
                 new IdGenerator()));
 
-    assertEquals("There should be one opportunity", 1, ops.size());
+    assertEquals("There should be two opportunities", 2, ops.size());
```
afd, line 74: Do:

```
assertEquals(0, pipelineInfo.getNumUniforms());
```

This should also pass, because the act of finding opportunities should not have in itself changed the shader job.

I believe that with your current implementation this assertion will fail, because the constructor of LiteralToUniformReductionOpportunity has side effects on the shader job.

(Please include these assertions in your next commit.)

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Removed these.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Moved to applyReductionImpl.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
amakin-siru: Added assertions.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
```diff
@@ -25,40 +26,277 @@
...
+    assertEquals(0, pipelineInfo.getNumUniforms());
+
+    ops.forEach(AbstractReductionOpportunity::applyReduction);
+
+    CompareAsts.assertEqualAsts(fragmentShaderReplaced, shaderJob.getFragmentShader().get());
+    CompareAsts.assertEqualAsts(vertexShaderReplaced, shaderJob.getVertexShader().get());
+
+    assertTrue(shaderJob.getPipelineInfo().hasUniform(Constants.FLOAT_LITERAL_UNIFORM_VALUES));
+    assertTrue(shaderJob.getPipelineInfo().hasUniform(Constants.INT_LITERAL_UNIFORM_VALUES));
+  }
```
paulthomson, line 301: I suggest adding a test where you apply just one reduction opportunity at first. Then check that just the one shader had the array uniform added, and the other shader stays the same (it does not have the array uniform). Then manually add another uniform to the pipeline info. Then find and apply the remaining opportunities in both shaders. Check that the bindings for the manually added uniform and the existing array uniform have not changed. I suggest this because I think this test (or a similar test) would currently fail.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,121 @@
...
+    // Adds the new array declaration to every translation unit.
+    for (TranslationUnit tu: shaderJob.getShaders()) {
+
+      // Declares the array if it didn't exist. Otherwise, removes it and adds a new declaration
+      // with increased size.
+      Optional<Declaration> oldDeclaration = Optional.empty();
+      final List<VariablesDeclaration> declarations = tu.getUniformDecls();
+      for (VariablesDeclaration declaration : declarations) {
+        if (declaration.getDeclInfo(0).getName().equals(arrayName)) {
+          oldDeclaration = Optional.of(declaration);
```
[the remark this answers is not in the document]
amakin-siru: A good remark. Removed these in the re-write.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,121 @@
...
+      arrayName = Constants.FLOAT_LITERAL_UNIFORM_VALUES;
+      basicType = BasicType.FLOAT;
+      numericValue = Float.valueOf(((FloatConstantExpr) literalExpr).getValue());
+    }
+
+    if (!shaderJob.getPipelineInfo().hasUniform(arrayName)) {
+      shaderJob.getPipelineInfo().addUniform(arrayName, basicType,
+          Optional.of(0), new ArrayList<>());
+      shaderJob.getPipelineInfo().addUniformBinding(arrayName, false,
+          shaderJob.getPipelineInfo().getNumUniforms());
```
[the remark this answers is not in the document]
amakin-siru: Fixed in the re-write.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,121 @@
...
+    if (literalExpr instanceof IntConstantExpr) {
+      arrayName = Constants.INT_LITERAL_UNIFORM_VALUES;
+      basicType = BasicType.INT;
+      numericValue = ((IntConstantExpr) literalExpr).getNumericValue();
+    } else {
+      arrayName = Constants.FLOAT_LITERAL_UNIFORM_VALUES;
+      basicType = BasicType.FLOAT;
+      numericValue = Float.valueOf(((FloatConstantExpr) literalExpr).getValue());
+    }
+
```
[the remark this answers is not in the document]
amakin-siru: Done.

reducer/src/test/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunitiesTest.java
amakin-siru: A new test case added accordingly to the suggestion.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -251,6 +251,37 @@ public void removeUniformBinding(String uniformName) {
...
+            .map(item -> item.getValue().getAsJsonObject()
+            .get("binding").getAsInt()).collect(Collectors.toList());
+
+    for (Integer binding: bindings) {
+      // If the result of subtraction is bigger than 1, an unused binding number was found.
+      if (binding - number > 1) {
+        break;
+      }
+      number++;
+    }
```
paulthomson, line 280: I am a bit worried about this method. 

E.g. bindings = [ 1 ]
result is: 1. This binding is used.

bindings = [ 1, 2 ]
result is: 2. This binding is used.

bindings = [ 0 ]
result is: 1. OK.

bindings = [ 0, 1 ]
result is: 2. OK

bindings = [ 2, 0 ]
result is: 0. This binding is used.

I think it would be easier and clearer to iterate from 0 to max int until you find a number that is not used (by checking a set of used binding).

amakin-siru: That's right. It fails if the numbers are in arbitrary order. Fixed now as suggested.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -251,6 +251,35 @@ public void removeUniformBinding(String uniformName) {
     }
   }
 
+  public int appendValueToUniform(String uniformName, Number value) {
```
afd, line 254: Can you add a JavaDoc comment for this method, in particular explaining what the returned value means?

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
```diff
@@ -184,6 +194,21 @@ public TranslationUnit clone() {
         .collect(Collectors.toList());
   }
 
+  public VariablesDeclaration getUniformDeclaration(String name) {
```
afd, line 197: Can you add a JavaDoc comment for this method, emphasising that a precondition of the method is that the uniform exists?

Also, can you fix the method so that it handles multiple and zero uniforms in the same variables declaration?

For example, in:

`uniform float a, b;`

it looks like your implementation would not currently find `b`.

I also wonder whether the method might crash on:

`uniform float;`

which is weird but legal.

Please make sure you add tests for both of these cases.

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
```diff
@@ -184,6 +194,21 @@ public TranslationUnit clone() {
...
+    return getGlobalVariablesDeclarations().stream()
+        .filter(item -> item.getBaseType().hasQualifier(TypeQualifier.UNIFORM)
+            && item.getDeclInfo(0).getName().equals(name))
+        .findAny()
+        .get();
+  }
+
+  public boolean hasUniformDeclaration(String name) {
```
afd, line 205: Similar comment to the above - it should handle multiple uniforms or zero uniforms in a single declaration (and you may as well add a JavaDoc comment.)

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
```diff
@@ -148,6 +148,16 @@ public void addDeclarationBefore(Declaration newDecl, Declaration existingDecl)
     throw new IllegalArgumentException("Existing declaration not found.");
   }
 
+  public void updateTopLevelDeclaration(Declaration newDecl, Declaration existingDecl) {
```
afd, line 151: As you have added some public methods to TranslationUnit, please add corresponding tests to TranslationUnitTest.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
afd: Also: please add tests to PipelineInfoTest for each of the public methods that you have added or changed.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -251,6 +251,35 @@ public void removeUniformBinding(String uniformName) {
...
+      throw new IllegalArgumentException("Uniform declaration not found.");
+    }
+
+    dictionary.getAsJsonObject(uniformName).get("args").getAsJsonArray().add(value);
+
+    return dictionary.getAsJsonObject(uniformName).get("args").getAsJsonArray().size() - 1;
+  }
+
+  public int getUnusedBindingNumber() {
+    int number = 0;
```
afd, line 266: No point declaring this here - clearer to declare it in the header of the `for` loop below.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -251,6 +251,35 @@ public void removeUniformBinding(String uniformName) {
...
+    List<Integer> bindings =
+        dictionary.entrySet().stream()
+            .filter(item -> item.getValue().getAsJsonObject().has("binding"))
+            .map(item -> item.getValue().getAsJsonObject()
+            .get("binding").getAsInt()).collect(Collectors.toList());
+
+    for (;number < Integer.MAX_VALUE; number++) {
+      if (bindings.contains(number)) {
+        break;
+      }
```
afd, line 277: This seems wrong - shouldn't you break if `bindings` does *not* contain `number`?

Also: instead of a break, it would be better to directly return `number`.

After the loop you could throw a new RuntimeException with a message saying that the code after the loop should be unreachable (as we should never have `MAX_VALUE` used bindings).

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,108 @@
...
 
   @Override
   void applyReductionImpl() {
-    // TODO: ...
+
+    final Number numericValue;
+    final String arrayName;
+    final BasicType basicType;
+
+    // If the uniform array doesn't exist in the pipeline info, adds an empty array.
```
afd, line 66: This comment looks like it should be moved to the next conditional block.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
```diff
@@ -36,7 +38,21 @@ private LiteralToUniformReductionOpportunities() {
 
     for (TranslationUnit tu : shaderJob.getShaders()) {
       new StandardVisitor() {
-        // TODO: something.
+
+        @Override
+        public void visitIntConstantExpr(IntConstantExpr intConstantExpr) {
```
afd, line 43: Is support for UintConstantExpr coming in the next PR?  (If so that's fine, but let's not forget.)

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,108 @@
...
+    // in every translation unit where the declaration is present.
+    final int arraySize = shaderJob.getPipelineInfo().getArgs(arrayName).size();
+    final ArrayInfo arrayInfo = new ArrayInfo(new IntConstantExpr(String.valueOf(arraySize)));
+    final VariableDeclInfo variableDeclInfo = new VariableDeclInfo(arrayName,
+        arrayInfo, null);
+    final VariablesDeclaration arrayDecl = new VariablesDeclaration(
+        new QualifiedType(basicType, Arrays.asList(TypeQualifier.UNIFORM)), variableDeclInfo
+    );
+
+    if (!this.translationUnit.hasUniformDeclaration(arrayName)) {
```
afd, line 103: Before this if I would have a comment: "In the case that this translation unit does not have a suitable uniform array declaration, we add one."

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,108 @@
...
+      this.translationUnit.addDeclaration(arrayDecl);
+    }
+
+    for (TranslationUnit tu: shaderJob.getShaders()) {
```
afd, line 107: I would comment before this loop: "We go through each translation unit in the shader job and update its existing array uniform to refer to this new array uniform."

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,108 @@
...
+      if (tu.hasUniformDeclaration(arrayName)) {
+        tu.updateTopLevelDeclaration(arrayDecl, tu.getUniformDeclaration(arrayName));
```
afd, line 109: Is the existing translation unit guaranteed to have the array uniform?

Example: suppose our shader job has a vertex shader and a fragment shader, and neither originally have the array uniform.

We then replace a literal in the fragment shader.

As a result, we will add the array uniform to the fragment shader (via the code above).  Then we will go through all the shaders and do this replacement - but there will be nothing to replace in the vertex shader.

Can you make sure we have a test for that case?

afd: If I am correct then I think you should replace the conditional that adds the array uniform if not already present and the loop that does replacement with a single loop that considers each translation unit in turn, and adds the array if not present or replaces it if present.

paulthomson: I am not so sure Ally. I think we should only ever _add_ the declaration to the target `tu`, if needed. We should loop through _all_ other `tu`s to make sure we update their size, if needed.

paulthomson: The current code _adds_ to the target `tu`. It updates every other `tu` but only if the declaration is present. It looks good to me.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
```diff
@@ -16,27 +16,108 @@
...
+    final int arraySize = shaderJob.getPipelineInfo().getArgs(arrayName).size();
+    final ArrayInfo arrayInfo = new ArrayInfo(new IntConstantExpr(String.valueOf(arraySize)));
+    final VariableDeclInfo variableDeclInfo = new VariableDeclInfo(arrayName,
+        arrayInfo, null);
+    final VariablesDeclaration arrayDecl = new VariablesDeclaration(
+        new QualifiedType(basicType, Arrays.asList(TypeQualifier.UNIFORM)), variableDeclInfo
+    );
+
+    if (!this.translationUnit.hasUniformDeclaration(arrayName)) {
+      this.translationUnit.addDeclaration(arrayDecl);
```
paulthomson, line 104: Could remove use of `this` throughout. There might be an inspection you can enable in IntelliJ that warns about unnecessary use of `this`.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
afd: Oh yes - you're absolutely right.  @amakin-siru - please ignore my suggestion.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
afd: Agreed - I would avoid `this` except in constructors or .equals methods.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
```diff
@@ -17,15 +17,78 @@
...
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 public final class LiteralToUniformReductionOpportunities {
 
   private LiteralToUniformReductionOpportunities() {
-    // This class just provides a static method.
+    // This class just provides a static methods.
```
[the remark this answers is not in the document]
amakin-siru: Fixed.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Fixed.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Changed accordingly.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Fixed.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
amakin-siru: Added the JavaDoc comment and the tests.

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
amakin-siru: Added the comment and the tests for both cases. The method seemed to work fine with zero uniforms in the variables declaration but didn't work with multiple uniforms. A fix to the latter case added.

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
amakin-siru: Done.

ast/src/main/java/com/graphicsfuzz/common/ast/TranslationUnit.java
amakin-siru: Done.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
amakin-siru: Moved the declaration to the header of the loop.

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
amakin-siru: Fixed the bug and changed as requested.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
amakin-siru: Support included in the new PR.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Done.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Unnecessary use of this keyword removed.

reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunity.java
amakin-siru: Works with multi-shader shader job now.

PR #997, opened 2020-06-03
Added --add-braces option to PrettyPrint

jarisiru:
This required moving of addBraces between packages


ast/src/test/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitorTest.java
```diff
@@ -580,4 +580,34 @@ public void testParseAndPrintNestedLoop() throws Exception {
...
+          "void main(void)\n"
+        + "{\n"
+        + "  if (condition) {\n"
+        + "  do_something();\n"
+        + "  }\n"
+        + "}\n";
+
+    final File fragFile = temporaryFolder.newFile("shader.frag");
+    final File outFile = temporaryFolder.newFile("out.frag");
+    FileUtils.writeStringToFile(fragFile, frag, StandardCharsets.UTF_8);
```
paulthomson, line 601: Minor thing: could you create an instance of ShaderJobFileOperations() and use that for all file operations. We don't do this consistently, but we should with new code. If it isn't available (because it isn't a dependency of this module) then don't worry.

jarisiru: ShaderJobFileOperations is not available in this scope (would cause circular dependency)

PR #998, opened 2020-06-03
Add demoChecker to run all demos and log return codes

nipunG314:
It runs all the demo binaries built in `experiments` directory and logs the resulting return codes in a file in the current working directory.

### DIFF
--- a/experiments/demoChecker.sh
+++ b/experiments/demoChecker.sh
@@ -1,0 +1,213 @@
+#!/bin/bash
+
+# Copyright 2020 The GraphicsFuzz Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sample Repo Paths
+#
+# This script is designed with the expectation that
+# it is being run in a directory where all the
+# corresponding build scripts have been run.
+#
+
+export VK_ICD_FILENAMES="$(pwd)/swiftshader/build-coverage/Linux/vk_swiftshader_icd.json"
+
+IMAGINATION_DEMOS=Imagination-Vulkan-Samples
+GOOGLE_SAMPLES=vulkan_test_applications
+KHRONOS_SAMPLES=Vulkan-Samples
+SASCHA_DEMOS=SaschaWillems_Vulkan
+
+# Default values
+OUTPUT_FILE_NAME="demoCheckerLog"
+TIMEOUT_DURATION=30
+TIMEOUT_KILL_DURATION=5
+
+SASCHA_FLAG=true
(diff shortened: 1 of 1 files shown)


experiments/scripts/demoChecker.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
```
hevrard, line 2: Please add a copyright + license header

experiments/scripts/demoChecker.sh
```diff
@@ -0,0 +1,78 @@
...
+# Default values
+INPUT_DIR_COUNT=0
+INPUT_DIRS=()
+OUTPUT_DIR=$HOME/.demoChecker
```
hevrard, line 6: Your default here is to create a hidden file in `$HOME`, I don't think it is a good idea because users would have a hard time to find it afterwards. I think a better default would be to create a "demoChecker" in the current directory:

```
OUTPUT_DIR=demoChecker
```

experiments/scripts/demoChecker.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# Run through the executables in the INPUT_DIRS
+for dir in "${INPUT_DIRS[@]}"
+do
+    echo "Dir: " $dir
+    pushd $dir
+    for file in $(find . -executable -type f)
+    do
+        echo "File: " $file
+        timeout -k $TIMEOUT_KILL_DURATION $TIMEOUT_DURATION $file
```
hevrard, line 68: I don't think this will work with the Khronos samples, which require the sample name as an argument?

experiments/scripts/demoChecker.sh
```diff
@@ -0,0 +1,78 @@
...
+        returnCode=$?
+        if [ $returnCode -eq 0 ] || [ $returnCode -eq 124 ] || [ $returnCode -eq 137 ]; then
```
hevrard, line 70: Hmmm, it is a bit tricky here.

Can we assume all demo apps run indefinitely, until we stop them?
If we can assume this, then the only "pass" case would be for the app to timeout: if the apps terminates earlier (even with code 0), then it means that it did not run properly.

In any case, it's a good idea to store the return code alongside the file name. So I would transform this into:

```
# We expect apps to timeout, any other behavior is a failure
if [ $returnCode -eq 124 ] || [ $returnCode -eq 137 ] ; then
    echo $file $returnCode >> "$OUTPUT_DIR/passFile"
else
    echo $file $returnCode >> "$OUTPUT_DIR/failFile"
fi
```

experiments/scripts/demoChecker.sh
```diff
@@ -0,0 +1,78 @@
...
+            TIMEOUT_DURATION="$2"
+            shift
+            shift
+        ;;
+        -tk|--timeout-kill-duration)
+            TIMEOUT_KILL_DURATION="$2"
+            shift
+            shift
+        ;;
+    esac
```
hevrard, line 42: Add a `--help` flag that print an help message about how to use this script.

experiments/scripts/demoChecker.sh
nipunG314: Added

experiments/scripts/demoChecker.sh
nipunG314: I think we can assume that. Out of all the demos I've seen so far, the only ones that finished early were the Vulkan Compute demos. And even they requried a final confirmation to fully terminate.

PR #1001, opened 2020-06-04
Added precision declaration reducer to reduction driver

jarisiru:


jarisiru:
Changed more than expected. Checking what went wrong.

PR #1005, opened 2020-06-05
Add build scripts for Vulkan Khronos Samples and Imagination Demos

nipunG314:
* Khronos Samples
* Imagination Demos

### DIFF
--- a/experiments/build-Khronos_Samples_Vulkan.sh
+++ b/experiments/build-Khronos_Samples_Vulkan.sh
@@ -14,16 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Build one of the Vulkan demos Repo
-# {Sascha Willems, Google Test Applications, Khronos Samples, Imagination Demos}
-#
-# Run from the top level directory of the repo
-# Requires that cmake and the other dependencies 
-# for the given repo have to be fulfiled
-#
-# In particular, for the Sascha Willems demos, run
-# the download_assets.py script to download the
-# required assets before running this.
+# Build the Khronos Samples Repo
 
 set -e
 set -x
--- a/experiments/build-Khronos_Samples_Vulkan.sh
+++ b/experiments/build-Khronos_Samples_Vulkan.sh
@@ -29,10 +29,13 @@ if [ -z "$VULKAN_SDK" ]; then
 fi
 
 #### Get the source
-git clone --recurse-submodules https://github.com/KhronosGroup/Vulkan-Samples.git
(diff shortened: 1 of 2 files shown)


experiments/build-Khronos_Samples_Vulkan.sh
```diff
@@ -0,0 +1,53 @@
...
+
+#### Check dependencies
+if [ -z "$VULKAN_SDK" ]; then
+    echo "VULKAN_SDK is empty, missing Vulkan SDK"
+    exit 1
+fi
+sudo apt-get install cmake g++ xorg-dev libglu1-mesa-dev
+
+#### Get the source
+git clone --recurse-submodules https://github.com/KhronosGroup/Vulkan-Samples.git
```
hevrard, line 42: For git submodules, it is best to update them after the checkout to a particular commit. Please update to:

```
git clone https://github.com/KhronosGroup/Vulkan-Samples.git

(
  cd .....
  git checkout .....
 
  # update submodules after checking out a particular commit
  git submodule update --init --recursive
)
```

experiments/build-Khronos_Samples_Vulkan.sh
```diff
@@ -0,0 +1,53 @@
...
+(
+    cd Vulkan-Samples
+    git checkout -b gsoc 8155762d9395de224e6ab1b22ae8e5880faabd3e
+)
+
+#### Build
+(
+    cd Vulkan-Samples
+    cmake -G "Unix Makefiles" -H. -Bbuild/linux -DCMAKE_BUILD_TYPE=Release
+    cmake --build build/linux --config Release --target vulkan_samples -- -j4
```
hevrard, line 52: Would it be possible to use Ninja rather than Makefiles here?

experiments/build-Khronos_Samples_Vulkan.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# This script download and builds the Khronos Vulkan Samples
+# https://github.com/KhronosGroup/Vulkan-Samples
+
+#### Check dependencies
+if [ -z "$VULKAN_SDK" ]; then
+    echo "VULKAN_SDK is empty, missing Vulkan SDK"
+    exit 1
+fi
+sudo apt-get install cmake g++ xorg-dev libglu1-mesa-dev
```
hevrard, line 39: You can assume apt dependencies are installed already.

We should avoid calling "sudo" in a build script, because it typically requires the user to enter a password.

nipunG314: Got it. I'll remove that line then.

PR #1006, opened 2020-06-08
Add find_uncovered_api.py

nipunG314:
It creates a list of the functions in swiftshader: libVulkan.cpp that have an execution count of 0
in the demo repos under the experiments directory


experiments/findUncoveredAPI.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
paulthomson, line 1: Nice. Can you change the bash scripts to use `#!/usr/bin/env bash` as well?

experiments/findUncoveredAPI.py
```diff
@@ -0,0 +1,45 @@
...
+import sys, json, subprocess
+
+json_string = subprocess.check_output(
+        "pushd swiftshader/build-coverage/src/Vulkan/CMakeFiles/vk_swiftshader.dir;\
+        gcov -i -t libVulkan.cpp.o;\
+        popd;\
+        exit 0",
+        shell=True,
+        executable="/bin/bash"
+).decode("utf-8").split('\n')[1]
```
paulthomson, line 26: I suggest not using `shell=True`. Instead, use the `cwd=` parameter to give the directory, and give the command using a list, like so:

```py
[shutil.which("gcov"), "-i", "-t", "libVulkan.cpp.o"]
```

I don't think you need to split the string; you can just use the entire string. Of course, this will fail if there are multiple JSON objects, but if there are multiple objects then something has gone wrong. You could add a comment to note that the output is one JSON object per line, but we only expect one object.

experiments/findUncoveredAPI.py
```diff
@@ -0,0 +1,45 @@
...
+
+libVulkanIndex = -1
+for index in range(len(files_array)):
+    if "libVulkan" in files_array[index]['file']:
+        libVulkanIndex = index
+        break
+
+if libVulkanIndex == -1:
+    print("libVulkan.cpp not found in gcov data")
+    sys.exit(0)
```
paulthomson, line 38: I would do exit(1) (or some number other than 0, because 0 means success)

experiments/findUncoveredAPI.py
```diff
@@ -0,0 +1,45 @@
...
+        executable="/bin/bash"
+).decode("utf-8").split('\n')[1]
+
+files_array = json.loads(json_string)['files']
+
+libVulkanIndex = -1
+for index in range(len(files_array)):
+    if "libVulkan" in files_array[index]['file']:
+        libVulkanIndex = index
+        break
```
paulthomson, line 34: This is probably not very Pythonic, but I am no expert and I think it is fine for now.

experiments/findUncoveredAPI.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+
+# Copyright 2020 The GraphicsFuzz Project Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
paulthomson, line 5: Oh one more comment: Could you name the file something like `find_uncovered_api.py`? We tend to use this style for Python files.

experiments/findUncoveredAPI.py
nipunG314: Sure. Should that be in this PR or a separate one?

experiments/findUncoveredAPI.py
nipunG314: I'll make the change to using `cwd=`.

The reason I need to move into the specified directory is because gcov expects that the .gcda file will be in the current working directory. If we pass in a relative directory path instead of just a file name, it would complain that it cannot open the notes file.

The reason I was splitting the string was that pushd and popd emit the current state of their stack to standard output. This would be included in json_string, which doesn't make any sense. So I needed to split the json_string in order to get a valid JSON string.

experiments/findUncoveredAPI.py
nipunG314: Would this be an improvement?

```python
files_array = json.loads(json_string)['files']
lib_vulkan_record = None

# We expect only one file_record to match
# libVulkan
for file_record in files_array:
    if "libVulkan" in file_record['file']:
        lib_vulkan_record = file_record

if lib_vulkan_record == None:
    print("libVulkan.cpp not found in gcov data")
    sys.exit(1)

functions = lib_vulkan_record['functions']
```

experiments/findUncoveredAPI.py
hevrard: Let's do it in a separate PR

experiments/find_uncovered_api.py
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys, json, subprocess
+
+json_string = subprocess.check_output(
+    ["gcov", "-i", "-t", "libVulkan.cpp.o"],
+    cwd="swiftshader/build-coverage/src/Vulkan/CMakeFiles/vk_swiftshader.dir"
+)
```
hevrard, line 22: This code is correct, but for readability I would prefer to clearly define the command and the working dir before calling subprocess.check_output(). Also, use long version of option flags for clarity, and  comment the reason why you need a particular directory:

```
cmd = ["gcov", "--json-format", "--stdout", "libVulkan.cpp.o"]
# Gcov must be called in a precise location to find the gcda file
workdir = "swiftshader/build-coverage/src/Vulkan/CMakeFiles/vk_swiftshader.dir"
json_string = subprocess.check_output(cmd, cwd=workdir)
```

experiments/findUncoveredAPI.py
paulthomson: Yeah looks good! Apparently, it is common to avoid `== None` and instead just do `if not lib_vulkan_record:`. Which is not exactly the same, but is preferred where possible, apparently.

nipunG314: Okay, I've made the changes. Do these suffice?

PR #1022, opened 2020-07-01
Literal to uniform

amakin-siru:


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
```diff
@@ -43,8 +45,14 @@ private LiteralToUniformReductionOpportunities() {
         @Override
         public void visitIntConstantExpr(IntConstantExpr intConstantExpr) {
           super.visitIntConstantExpr(intConstantExpr);
-          opportunities.add(new LiteralToUniformReductionOpportunity(intConstantExpr, tu,
-              shaderJob, getVistitationDepth()));
+          final IParentMap parentMap = IParentMap.createParentMap(tu);
```
afd, line 48: I suggest that you instead make parentMap a private final field of the class and initialize it in the constructor.  Otherwise we traverse the whole AST every time we process an index expression, which gives  can get us into O(N^2) territory if a shader has loads of array indexing operations.

afd: Actually - I suggest doing this a different way entirely - I'll write a top-level review comment.

PR #1024, opened 2020-07-02
Added support for samplers in shader job generation

jarisiru:
Also added checks that uniforms and samplers with the same names don't exist, and that uniforms of different types do not have the same name.


common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -74,17 +74,24 @@ public PipelineInfo(String string) {
...
-    JsonObject info = new JsonObject();
-    info.addProperty("func", PipelineInfo.get(basicType, arrayCount.isPresent()));
-    JsonArray jsonValues = new JsonArray();
-    for (Number n : values) {
-      jsonValues.add(n);
-    }
-    info.add("args", jsonValues);
-    if (arrayCount.isPresent()) {
-      info.addProperty("count", arrayCount.get());
+    if (!dictionary.has(name)) {
```
afd, line 77: I'm keen on the style of eliminating all erroneous cases first, to avoid nesting (avoiding the "pyramid of doom").  So I would prefer you to do:

if (dictionary.has(name)) {
   // throw the exception
}
// No else needed; do the non-exceptional case

afd: I guess I'm wrong here, though, since in the "else" you don't unconditionally throw an exception.

In that case I suggest you use "else if", to avoid "else { if { "

common/src/main/java/com/graphicsfuzz/common/util/FragmentShaderToShaderJob.java
```diff
@@ -86,28 +87,39 @@ public static ShaderJob createShaderJob(TranslationUnit tu, IRandom generator) {
     // Iterate through all uniforms in 'tu'.  For each, add an entry to 'pipelineInfo' with a
     // randomized value (using 'generator' as the source of randomness).
```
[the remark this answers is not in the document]
jarisiru: Done

common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
jarisiru: I swapped them around anyway, and added clarifying comments

jarisiru:
Superceded by 1066

PR #1031, opened 2020-07-09
Fixes the literal-to-uniform feature so that the indices of the uniforms won't be replaced recursively.

amakin-siru:


reducer/src/main/java/com/graphicsfuzz/reducer/reductionopportunities/LiteralToUniformReductionOpportunities.java
```diff
@@ -41,6 +44,18 @@ private LiteralToUniformReductionOpportunities() {
...
+        public void visitArrayIndexExpr(ArrayIndexExpr arrayIndexExpr) {
+          // This prevents replacing the uniforms recursively. E.g. _GLF_int_values[0] does not
+          // become _GLF_int_values[_GLF_int_values[1]].
+          String name = ((VariableIdentifierExpr) arrayIndexExpr.getArray()).getName();
+          if (!name.equals(Constants.FLOAT_LITERAL_UNIFORM_VALUES)
+              && !name.equals(Constants.INT_LITERAL_UNIFORM_VALUES)
+              && !name.equals(Constants.UINT_LITERAL_UNIFORM_VALUES)) {
+            super.visitArrayIndexExpr(arrayIndexExpr);
+          }
+        }
```
afd, line 57: The trouble with this is that it will fail if the expression being indexed into is not a VariableIdentifierExpr.

Here's a shader on which it will probably fail:

```
#version 320 es

void main() {

  int A[3] = int[3](1, 2, 3);
  int B[3] = int[3](4, 5, 6);
  int x = (true ? A : B)[2];
 
}
```

Can you check that and if so add a test for it?

A pragmatic fix is:

if (arrayIndexExpr instanceof VariableIdentifierExpr) {
  String name = ...;
  if (name is one of the things we want to skip) {
    return;
  }
}
super.visit...;

-

amakin-siru: Fixed.

PR #1032, opened 2020-07-09
Update Java 11 dependencies

Vasniktel:
`javax.annotation` was removed in Java 11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@
 *.iml
 /.idea/
 
+# VSCode
+.vscode/
+
 # Maven.
 target/
 dependency-reduced-pom.xml
--- a/server-thrift-gen/pom.xml
+++ b/server-thrift-gen/pom.xml
@@ -64,6 +64,13 @@ limitations under the License.
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <dependency>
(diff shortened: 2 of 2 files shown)


afd:
@paulthomson could you review please?

paulthomson:
Thanks for this. This works but I don't understand why! The `javax.annotation.Generated` class no longer exists in JDK 11. The dependency added to the pom.xml file allows compilation to succeed, but the class is still not included in the output jars. And yet things still seem to run fine.

I have an alternative PR that prevents the thirft compiler outputting Java code with `@javax.annotation.Generated`. I may submit it tomorrow.

Vasniktel:
I think I know why the compilation succeeds. It has something to do with `<scope>provided</scope>` in the dependency's xml. When I tried using it without the scope, it failed. I think what it means is that the scope tells the mvn that the library will be 'provided' 
(not sure when and where) so it doesn't produce the compilation error. That's just a guess.

paulthomson:
Yeah, my confusion was: why don't our Java tools (e.g. Reducer) crash at runtime? They should fail to find the annotation. 

I figured it out. The `Generated` annotation is `@Retention(SOURCE)`, which means it is discarded at build time. Thus, the final class does not contain the annotation.

PR #1042, opened 2020-07-29
gfauto: require TARGET_ENV

paulthomson:
...when extracting SPIRV-ASM shaders from an AmberScript file.


paulthomson:
We should not merge this until all existing dEQP tests have been fixed.

PR #1049, opened 2020-08-18
Fix the halving of the granularity at which reduction opportunities a…

amakin-siru:
…re applied.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
@@ -402,34 +402,7 @@ public final class PipelineInfo {
 
   @Override
   public PipelineInfo clone() {
-
-    final PipelineInfo pipelineinfo = new PipelineInfo(toString());
-
-    // These are copied separately since constructing PipelineInfo from a string converts
-    // Number objects into LazilyParsedNumber objects.
-    String[] names = {Constants.FLOAT_LITERAL_UNIFORM_VALUES, Constants.INT_LITERAL_UNIFORM_VALUES,
-        Constants.UINT_LITERAL_UNIFORM_VALUES};
-    BasicType[] types = {BasicType.FLOAT, BasicType.INT, BasicType.UINT};
-
-    for (int i = 0; i < names.length; i++) {
-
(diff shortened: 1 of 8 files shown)


afd (changes requested):
Are the changes "clone" related to the changes to the granularity in the reducer?  If not then it would be better to do them as separate PRs.

I am not sure what problem the granularity change in the reducer is trying to fix.  However, the change does not look correct.

This is how the reduction algorithm is intended to work, before your change:

The reducer tries to remove a chunk of size |granularity| starting at |index|.  If that works, it tries again (because the opportunities starting at |index| will now be the ones that previously started at |index|+|granularity|.  If it doesn't work, the reducer increases |index| by |granularity|.

If we end up having |index| being larger than the number of opportunities then we have completed a pass, in chunks of size |granularity|, over all the reduction opportunities, so we set |index| to zero and do a new pass, but this time with chunks of size |granularity|/2.  Thus we only want to halve |granularity| when it's time to reset |index|.

afd (changes requested):
Thanks for the clever test case exposing the bug!  I'm going to attach a patch (to be applied on top of master) that fixes the underlying problem in a different way.

afd:
The reducer contained this code, in `SystematicReductionPassManager.cpp`:

```
      passIndex++;
      if (passIndex < currentPasses.size()) {
        anotherRoundWorthwhile |= !getCurrentPass().reachedMinimumGranularity();
      } else if (anotherRoundWorthwhile) {
        ...
```

The `anotherRoundWorthwhile` field is initialized to false and is meant to become true if some reduction pass in the current round of passes could have gone further.

The bug is that `passIndex` gets incremented before this check, so if the current pass is the *final* pass, there isn't a chance to set `anotherRoundWorthwhile` to true due to that pass not having reached its minimum granularity.

This lay hidden because - with many passes in play - it is unlikely to be a problem.  But with just one pass it's very likely to crop up!

Here is a patch that includes your very nice test, but fixes the problem a different way to the approach you took.  I think it's a better fix because it addresses the buggy logic directly.

Thanks for finding this bug!

[patch.txt](https://github.com/google/graphicsfuzz/files/5130987/patch.txt)

afd:
Closing this in favour of #1057.

PR #1052, opened 2020-08-26
Fixed the clone method in PipelineInfo to preserve the types of the v…

amakin-siru:
…alues in uniform arrays.

### DIFF
--- a/common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
+++ b/common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
@@ -402,7 +402,34 @@ public final class PipelineInfo {
 
   @Override
   public PipelineInfo clone() {
-    return new PipelineInfo(toString());
+
+    final PipelineInfo pipelineinfo = new PipelineInfo(toString());
+
+    // These are copied separately since constructing a PipelineInfo object from a string converts
+    // Number objects into LazilyParsedNumber objects.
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/graphicsfuzz/common/util/PipelineInfo.java
```diff
@@ -402,7 +402,34 @@ private JsonObject lookupUniform(String uniformName) {
 
   @Override
   public PipelineInfo clone() {
-    return new PipelineInfo(toString());
+
+    final PipelineInfo pipelineinfo = new PipelineInfo(toString());
```
afd, line 406: I suggest calling this `result` - I like to follow a convention where the thing to ultimately be returned is called `result`.  (I wouldn't suggest this minor change if I wasn't requesting other changes :))

afd (changes requested):
Is the new test meant to expose the problem that this PR fixes?

I checked out your branch and reverted the implementation to clone to the old one, and then ran the new test, and it passed.  Thus either the new test doesn't expose the problem, or there is something platform-dependent going on here.

In any case, though, I don't understand why it would be necessary to treat these particular uniforms specially.  If there's a problem relating to lazily-parsed numbers then why doesn't it apply to *all* uniforms - why only to the uniforms named things like "GLF_uniform_int_values"?

Seeing a failing test would really help me to understand what the underlying problem is here.

afd:
Closing this as it is superseded by #1056

Issue #1054, opened 2020-08-27
Cloning of a PipelineInfo object doesn't preserve the types of the uniform values


amakin-siru:
Literal-to-uniform pass doesn't work right since it adds the same literal multiple times into a PipelineInfo object if a translation unit includes more than one occurrence of the literal.

The reason for this behaviour is that the clone method in PipelineInfo converts the types of the uniform values into LazilyParsedNumbers and this causes the comparison in `LiteralToUniformReductionOpportunity.applyReductionImpl(...)` to fail since it compares a Number object to LazilyParsedNumber objects.

PR #1055, opened 2020-08-27
Add test to expose bug in PipelineInfo

amakin-siru:
… pipelineinfo object has been cloned.

TODO(#1054)


afd (approved):
Thanks for the test.

PR #1060, opened 2020-08-28
Fix LLVM signature recognition

afd:
Recent testing using LLPC shows that the format of fatal errors has
changed slightly.  This PR relaxes the regex used to recognize fatal
errors to tolerate their old form and their new form.  The same
relaxation is applied for machine code errors to account for the
possibility of an analogous formatting change being applied there.
Some tests are added related to this.


paulthomson (approved):
LGTM!

PR #1062, opened 2020-09-07
Use defines for literal-uniform array elements

amakin-siru:
…x into them instead.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
@@ -570,36 +570,40 @@ public class PrettyPrinterVisitor extends StandardVisitor {
     visit(ternaryExpr.getElseExpr());
   }
 
+  private boolean maybeEmitKnownUniformMacro(ArrayIndexExpr arrayIndexExpr,
+                                           String knownUniformArrayName, String prefix) {
+    /* If arrayIndexExpr indexes into a literal-to-uniform array, output a macro name instead. */
+    if (arrayIndexExpr.getArray() instanceof VariableIdentifierExpr
+        && ((VariableIdentifierExpr) arrayIndexExpr.getArray()).getName()
+        .contains(knownUniformArrayName)
+        && uniformValues.isPresent()
+        && uniformValues.get().getValues(knownUniformArrayName).isPresent()) {
+
+      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
+      List<String> values = uniformValues.get().getValues(knownUniformArrayName).get();
+      out.append("_").append(prefix).append("_").append(values.get(index).replace(".", "_"));
+      return true;
+    }
+    return false;
+  }
+
   @Override
   public void visitArrayIndexExpr(ArrayIndexExpr arrayIndexExpr) {
-    /* If arrayIndexExpr indexes into a literal-to-uniform array, output a #define instead. */
-    if (arrayIndexExpr.getArray().getText().contains(Constants.INT_LITERAL_UNIFORM_VALUES)
-        && uniformValues.isPresent()
-        && uniformValues.get().getValues(Constants.INT_LITERAL_UNIFORM_VALUES).isPresent()) {
-      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
-      List<String> values =
-          uniformValues.get().getValues(Constants.INT_LITERAL_UNIFORM_VALUES).get();
-      out.append("_int_" + values.get(index));
-    } else if (arrayIndexExpr.getArray().getText().contains(Constants.UINT_LITERAL_UNIFORM_VALUES)
-        && uniformValues.isPresent()
-        && uniformValues.get().getValues(Constants.UINT_LITERAL_UNIFORM_VALUES).isPresent()) {
-      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
-      List<String> values =
-          uniformValues.get().getValues(Constants.UINT_LITERAL_UNIFORM_VALUES).get();
-      out.append("_uint_" + values.get(index));
-    } else if (arrayIndexExpr.getArray().getText().contains(Constants.FLOAT_LITERAL_UNIFORM_VALUES)
-        && uniformValues.isPresent()
-        && uniformValues.get().getValues(Constants.FLOAT_LITERAL_UNIFORM_VALUES).isPresent()) {
-      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
-      List<String> values =
-          uniformValues.get().getValues(Constants.FLOAT_LITERAL_UNIFORM_VALUES).get();
-      out.append("_float_" + values.get(index).replace(".", "_"));
-    } else {
-      visit(arrayIndexExpr.getArray());
-      out.append("[");
-      visit(arrayIndexExpr.getIndex());
-      out.append("]");
+    if (maybeEmitKnownUniformMacro(arrayIndexExpr, Constants.INT_LITERAL_UNIFORM_VALUES, "int")) {
+      return;
     }
+    if (maybeEmitKnownUniformMacro(arrayIndexExpr, Constants.UINT_LITERAL_UNIFORM_VALUES, "uint")) {
+      return;
+    }
+    if (maybeEmitKnownUniformMacro(arrayIndexExpr, Constants.FLOAT_LITERAL_UNIFORM_VALUES,
+        "float")) {
+      return;
+    }
+
+    visit(arrayIndexExpr.getArray());
+    out.append("[");
+    visit(arrayIndexExpr.getIndex());
+    out.append("]");
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -572,10 +572,34 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
 
   @Override
   public void visitArrayIndexExpr(ArrayIndexExpr arrayIndexExpr) {
-    visit(arrayIndexExpr.getArray());
-    out.append("[");
-    visit(arrayIndexExpr.getIndex());
-    out.append("]");
+    /* If arrayIndexExpr indexes into a literal-to-uniform array, output a #define instead. */
+    if (arrayIndexExpr.getArray().getText().contains(Constants.INT_LITERAL_UNIFORM_VALUES)
```
afd, line 576: Instead of asking whether the text of the arrayIndexExpr contains the string, check that the expression is a variable identifier expression and that, if so, it's name is .equals to the string:

```
if (arrayIndexExpr instanceof VariableIdentiferExpr && ((VariableIdentifierExpr) arrayIndexExpr).getName().equals(...)
```

What you did with checking substrings might work in practice but it's a bit fragile.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -572,10 +572,34 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
...
+        && uniformValues.isPresent()
+        && uniformValues.get().getValues(Constants.INT_LITERAL_UNIFORM_VALUES).isPresent()) {
+      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
+      List<String> values =
```
afd, line 580: Make these final - always make local variables final if you can.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -572,10 +572,34 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
...
+          uniformValues.get().getValues(Constants.FLOAT_LITERAL_UNIFORM_VALUES).get();
+      out.append("_float_" + values.get(index).replace(".", "_"));
+    } else {
+      visit(arrayIndexExpr.getArray());
+      out.append("[");
+      visit(arrayIndexExpr.getIndex());
+      out.append("]");
+    }
```
afd, line 602: Is there a way to avoid all this duplication?

How about a function:

```
private bool maybeEmitKnownUniformMacro(ArrayIndexExpr arrayIndexExpr, String knownUniformArrayName, String prefix);
```

This function could check whether the array index expr's array matches the array name, in which case it could emit the symbol and return true.  Otherwise it could return false.

Then you could do something like:

```
if (maybeEmitKnownUnformMacro(arrayIndexExpr, Constants.INT_LITERAL_UNIFORM_VALUES, "int") {
  return;
}
if (maybeEmitKnownUnformMacro(arrayIndexExpr, Constants.UINT_LITERAL_UNIFORM_VALUES, "uint") {
  return;
}
...
// The original code
```

common/src/main/java/com/graphicsfuzz/common/util/PipelineUniformValueSupplier.java
```diff
@@ -30,6 +30,10 @@ public PipelineUniformValueSupplier(PipelineInfo pipelineInfo) {
 
   @Override
   public Optional<List<String>> getValues(String name) {
-    return Optional.of(this.pipelineInfo.getArgs(name));
+    if (pipelineInfo.hasUniform(name)) {
+      return Optional.of(this.pipelineInfo.getArgs(name));
+    } else {
+      return Optional.empty();
+    }
```
afd, line 37: It's a matter of taste really, but the "else" is redundant; you can always rewrite:

```
if (e) {
  return f;
} else {
  return g;
}
```

as

```
if (e) {
  return f;
}
return g;
```

I slightly prefer the latter.

afd (changes requested):
This looks good!  Some refactoring requested to avoid duplication but otherwise good to go.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -754,6 +778,41 @@ public void visitTranslationUnit(TranslationUnit translationUnit) {
       emitGraphicsFuzzDefines(out, translationUnit.getShadingLanguageVersion());
     }
 
+    // Emit a #define for each value in the literal-to-uniform arrays.
+    if (uniformValues.isPresent()) {
+      if (uniformValues.get().getValues(Constants.INT_LITERAL_UNIFORM_VALUES).isPresent()) {
+        List<String> intValues =
```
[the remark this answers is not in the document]
amakin-siru: All the added local variables are final now.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -754,6 +778,41 @@ public void visitTranslationUnit(TranslationUnit translationUnit) {
...
+            uniformValues.get().getValues(Constants.INT_LITERAL_UNIFORM_VALUES).get();
+        int index = 0;
+        for (String value : intValues) {
+          out.append("#define _int_" + value + " " + Constants.INT_LITERAL_UNIFORM_VALUES
+              + "[" + index + "]" + newLine());
+          index++;
+        }
+      }
```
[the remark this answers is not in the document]
amakin-siru: Done.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -572,10 +572,34 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
 
   @Override
   public void visitArrayIndexExpr(ArrayIndexExpr arrayIndexExpr) {
-    visit(arrayIndexExpr.getArray());
-    out.append("[");
-    visit(arrayIndexExpr.getIndex());
-    out.append("]");
+    /* If arrayIndexExpr indexes into a literal-to-uniform array, output a #define instead. */
```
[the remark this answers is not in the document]
amakin-siru: Replaced with more accurate word.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
amakin-siru: Changed.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
amakin-siru: I added the method and cleaned up the duplication.

common/src/main/java/com/graphicsfuzz/common/util/PipelineUniformValueSupplier.java
amakin-siru: Rewritten.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -570,8 +570,36 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
     visit(ternaryExpr.getElseExpr());
   }
 
+  private boolean maybeEmitKnownUniformMacro(ArrayIndexExpr arrayIndexExpr,
+                                           String knownUniformArrayName, String prefix) {
+    /* If arrayIndexExpr indexes into a literal-to-uniform array, output a macro name instead. */
```
afd, line 575: FYI: We prefer line comments (//) in this project, unless there's a good reason not to use them (no action required as it's not mandatory).

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -570,8 +570,36 @@ public void visitTernaryExpr(TernaryExpr ternaryExpr) {
...
+    if (arrayIndexExpr.getArray() instanceof VariableIdentifierExpr
+        && ((VariableIdentifierExpr) arrayIndexExpr.getArray()).getName()
+        .contains(knownUniformArrayName)
+        && uniformValues.isPresent()
+        && uniformValues.get().getValues(knownUniformArrayName).isPresent()) {
+
+      int index = ((IntConstantExpr) arrayIndexExpr.getIndex()).getNumericValue();
+      List<String> values = uniformValues.get().getValues(knownUniformArrayName).get();
```
afd, line 583: FYI, as per previous review I would prefer these to be final.

afd (approved):
This is great, thanks!  A couple of remarks FYI, but I'll merge (if the CI has passed).

PR #1064, opened 2020-09-10
Add signature recognition for AddressSanitizer

afd:
Adds basic recognition of AddressSanitizer signatures to gfauto, and
adds a test for assertion-related signatures.


paulthomson (approved):
LGTM!

PR #1065, opened 2020-09-10
Avoid string concatenation in pretty printer

afd:
Improves the pretty printer by writing straight to a print stream
rather than using many intermediate strings.

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
@@ -138,7 +138,7 @@ public class PrettyPrinterVisitor extends StandardVisitor {
   /**
    * Returns, via pretty printing, a string representation of the given node.
    *
-   * @param node          Node for which string representation is required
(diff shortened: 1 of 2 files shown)


ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
```diff
@@ -138,7 +138,7 @@ public static String prettyPrintAsString(IAstNode node) {
   /**
    * Returns, via pretty printing, a string representation of the given node.
    *
-   * @param node Node for which string representation is required
+   * @param node          Node for which string representation is required
```
paulthomson, line 141: Accidental change?

paulthomson (approved):
LGTM. Could make `newLine()` emit a new line I guess, similar to `append()`, but it seems fine either way.

ast/src/main/java/com/graphicsfuzz/common/tool/PrettyPrinterVisitor.java
afd: Good spot, thanks!

PR #1067, opened 2020-09-10
Always initialize variables during donation

### LINKED ISSUE #1063: Uninitialized vec4 variable
Graphicsfuzz has generated uninitialized vec4 variable `mediump vec4 GLF_live16gl_FragCoord;` into a variant shader. Zip folder contains the reference, reduced variant and unreduced variant shaders.
[shaders.zip](https://github.com/google/graphicsfuzz/files/5199873/shaders.zip)

Derived from: stable_collatz

afd:
To reduce the chances of undefined behaviour, this change forces all
variables to be initialized before a module is donated.

An existing test is fixed in the process; its brokenness was exposed
by this change.

Fixes #1063.


paulthomson (approved):
LGTM

PR #1070, opened 2020-09-16
gfauto: Add support for samplers

jarisiru:
This change adds texture sampler support in the fuzzed
shaders. It also generates a texture in the amber file
to be used. No mutation of texture parameters is
performed by this change.


gfauto/gfauto/amber_converter.py
```diff
@@ -355,12 +363,17 @@ def amberscript_uniform_buffer_def(uniform_json_contents: str, prefix: str) -> s
             raise ValueError("Error: unknown uniform type for function: " + func)
         uniform_type = uniform_types[func]
```
paulthomson, line 364: Please add an example to the above docstring for a sampler.

gfauto/gfauto/amber_converter.py
```diff
@@ -260,7 +260,13 @@ def amberscript_uniform_buffer_bind(uniform_json: str, prefix: str) -> str:
     for name, entry in uniforms.items():
         if name.startswith("$"):
             continue
-        if "binding" in entry.keys():
+        if entry["func"] in ["sampler2D", "sampler3D"]:
+            assert "texture" in entry.keys()
+            if entry["texture"] == "DEFAULT":
+                result += f"  BIND BUFFER default_texture AS combined_image_sampler SAMPLER {prefix}_{name} DESCRIPTOR_SET 0 BINDING 0\n"
```
paulthomson, line 266: I think there should always be a binding field, and we should use that instead of 0.

paulthomson (commented):
Looks good! I left a few comments.

gfauto/gfauto/amber_converter.py
```diff
@@ -260,7 +260,13 @@ def amberscript_uniform_buffer_bind(uniform_json: str, prefix: str) -> str:
     for name, entry in uniforms.items():
         if name.startswith("$"):
```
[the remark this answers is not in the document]
jarisiru: Done

gfauto/gfauto/amber_converter.py
jarisiru: Done

gfauto/gfauto/amber_converter.py
jarisiru: Done

paulthomson (approved):
LGTM! Thanks!

PR #1072, opened 2020-09-24
Add collection of stable shaders using samplers

jarisiru:
The samplers range from very simple texture reads to
shaders that read textures based on texture reads, and
various UV coordinate transformations.


afd (approved):
Thanks for this contribution!

PR #1077, opened 2020-09-26
Handle setting and pruning of sample uniforms

### LINKED ISSUE #415: Handle sampler uniforms
Currently sampler uniforms are ignored when pipeline state is populated.  Depending on whether we decide to properly support textures, we should think about handling these uniforms properly.

afd:
Fixes #415.


paulthomson (approved):
LGTM, except remove `generator/comp.spv`

afd:
Thanks - I have removed that file, well spotted.

PR #1079, opened 2020-10-01
Add shaders utilizing logic operations

jarisiru:


afd (approved):
Thanks for these shaders!

PR #1080, opened 2020-10-07
Signal discard from vertex to fragment shader

jarisiru:
In converted shader jobs, signal 'discard' from vertex to fragment
shader. This helps somewhat, but does not make the functionality
identical, as the shader execution is not terminated at discard.


afd (approved):
Looks great!

PR #1082, opened 2020-10-14
Enable comma operator fuzzing, cleanup

jarisiru:
This change enables the disabled comma operator fuzzing and
also removes the unused code that could be used to disable
the use of non-rectangular matrices.


afd (approved):
LGTM.

PR #1083, opened 2020-10-19
gfauto: improve Android permissions

paulthomson:
Run Amber to create the Amber cache directory on /sdcard. This might be necessary on newer Android versions to ensure the cache directory is accessible by the Amber app.


afd (approved):
LGTM.

PR #1084, opened 2020-10-19
gfauto: support SPIR-V donors and references

paulthomson:
Adds support for different donors and references when using SPIR-V fuzz. These should be placed in spirv_fuzz_donors/ and spirv_fuzz_references/, respectively.


afd (approved):
LGTM.

PR #1085, opened 2020-10-21
gfauto: add bucket_via_transformations

paulthomson:
By default, puts spirv-fuzz bad image bugs in buckets classified by the important transformation types that remain.


afd (approved):
Looks excellent!

PR #1090, opened 2020-11-04
Add linear algebra chessboard sample shader

andreperezmaselco:
This PR implements a sample shader that uses linear algebra instructions.

linear_algebra_chessboard


afd (approved):
Thanks for this contribution - sorry it took me so long to get to it!

PR #1097, opened 2020-11-14
gfauto: add process_seeds.py

paulthomson:
Processes log files and a list of seeds to output the crash signatures obtained by chunks of seeds.

### DIFF
--- a/gfauto/gfautotests/test_signature_util.py
+++ b/gfauto/gfautotests/test_signature_util.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 from pathlib import Path
+
 from gfauto import signature_util
 
 # Disable spell-checking for this file.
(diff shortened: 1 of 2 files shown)


gfauto/gfautotests/test_signature_util.py
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 from pathlib import Path
+
```
afd, line 18: Deliberate?

afd (changes requested):
Approved with one comment.

gfauto/gfautotests/test_signature_util.py
paulthomson: Yeah actually this is correct, but not currently picked up by the linter.

afd (approved):

PR #1100, opened 2020-11-15
Improvements to seed processing script

afd:
The script now works for spirv-opt and LLPC, and has the option to
output just numbers of unique signatures, which is useful for
gathering statistics to put in a spreadsheet.

### DIFF
--- a/gfauto/gfauto/process_seeds.py
+++ b/gfauto/gfauto/process_seeds.py
@@ -28,11 +28,12 @@ from gfauto.util import check
 
 
 def process_chunk(  # pylint: disable=too-many-locals;
-    chunk_num: int, chunk: Set[str], log_files: List[Path], output_file: TextIO
+    chunk_num: int, chunk: Set[str], log_files: List[Path], only_show_num_signatures: bool, look_for_spirv_opt_errors: bool, look_for_llpc_errors: bool, output_file: TextIO
 ) -> None:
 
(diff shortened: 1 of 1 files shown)


gfauto/gfauto/process_seeds.py
```diff
@@ -50,16 +51,33 @@ def process_chunk(  # pylint: disable=too-many-locals;
...
-                    continue
-                found_bug = True
-                end_line = i + 1
-                break
+                if look_for_spirv_opt_errors:
+                    match = re.match(r"Command failed: \['/usr/bin/catchsegv', .*spirv-opt'", line)
+                    if match:
+                        found_bug = True
+                        start_line = i + 1
+                        # TODO: Is there a reliable way to find the end of a spirv-opt failure?
```
paulthomson, line 59: Maybe:
 - Don't break once a bug is found.
 - if found_bug, do something different: 
    - Just look for "Exec (verbose): ..." to find the next "reduce" command, set the end line based on this and then break.
   - if we find "gfauto_interestingness_test: finished ..." then the output above this is actually the output from part of a reduction, so something has gone wrong.
   - if we don't find anything else and get to the end of the file, this is also fine. It just means we did not do a reduction and the file immediately ends after the "Command failed:" output.

paulthomson (approved):
Approved, with a suggestion for the TODO.

PR #1108, opened 2020-11-27
cov: add function coverage option

karineek:
I added a new option: --gcov_functions to generate instead of line coverage, the function coverage when using JSON format. I did testing to check that without the new flag, the tool still generates line coverage, and with the new flag, the tool generates function coverage.

### DIFF
--- a/gfauto/gfauto/cov_from_gcov.py
+++ b/gfauto/gfauto/cov_from_gcov.py
@@ -84,6 +84,9 @@ def main() -> None:
     if not parsed_args.gcov_path:
         parser.error("Please provide gcov_path")
 
+    if parsed_args.gcov_functions and not parsed_args.gcov_uses_json:
+        parser.error("Function coverage requires using --gcov_uses_json with gcc 9+.")
+
     gcov_path: str = parsed_args.gcov_path
 
     build_dir = parsed_args.build_dir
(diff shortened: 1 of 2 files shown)


karineek:
@googlebot I signed it!

afd:
@karineek I think you need to fix some formatting errors, using the check_all.sh and fix_all.sh scripts under gfauto.  Can you give that a try?  See "## Presubmit checks and protobuf generation" in the README.

gfauto/gfauto/cov_from_gcov.py
```diff
@@ -87,12 +94,21 @@ def main() -> None:
...
+        if parsed_args.gcov_functions
+        else ["lines", "line_number", "count"]
+    )
+
     data = cov_util.GetLineCountsData(
         gcov_path=gcov_path,
-        gcov_uses_json_output=parsed_args.gcov_uses_json,
+        gcov_uses_json_output=True
+        if parsed_args.gcov_functions
+        else parsed_args.gcov_uses_json,
```
afd, line 107: I think it would be better to pass parsed_args.gcov_uses_json here, and to check earlier whether this is inconsistent with parsed_args.gcov_functions, quitting with an error if there is an inconsistency.

afd (changes requested):
Thanks for the contribution, Karine!  I'd like @paulthomson to do a review, but I have one comment.

gfauto/gfauto/cov_from_gcov.py
karineek: I remove the "if" and added a check with an error message instead.

paulthomson (approved):

PR #1114, opened 2021-01-27
Avoid discard in live-donated code

### LINKED ISSUE #1110: gfauto reduction not semantically same
Few Graphicsfuzz bad image reduction results have semantically different reference and variant shaders.

Code block bellow is the diffrence between the reference and variant shaders. The switch case in the function `GLF_live5patternize` always ends up into case 2 that has discard in it. Whenever the function is called, it discards the pixel. The shaders are no longer semantically same.

```
float GLF_live5patternize(vec2 GLF_live5uv)
{
 // 7.3 mod 5.0 typecasted to int equals 2
 switch(int(mod((vec4(1.0, 7.3, 1.0, 1.0)).y, 5.0)))
  {
   case 0:
   return mix(pow(vec2(1.0).x, vec2(1.0).y), vec2(1.0).x, vec2(1.0).y);
   break;
   case 1:
   return mix(pow(vec2(1.0).y, vec2(1.0).y), vec2(1.0).y, vec2(1.0).x);
   break;
   case 2:
   discard;
   break;
   case 3:
   return mix(pow(vec2(1.0).y, vec2(1.0).y), vec2(1.0).y, vec2(1.0).y);
   break;
   case 4:
   return mix(pow(vec2(1.0).y, vec2(1.0).y), vec2(1.0).x, vec2(1.0).x);
   break;
  }
}

...

vec3 GLF_live5color = vec3(GLF_live5patternize(vec2(1.0)));
```
Drive file name: Graphicsfuzz_discard.zip

afd:
Removes discard statements from a donor during live code donation, instead of only removing discards from the statements that get locally donated.  This avoids roblems where a donated statement calls a function that contains a discard.

Some tests are also refactored.

Fixes #1110.

### DIFF
--- a/build/travis/check_headers.py
+++ b/build/travis/check_headers.py
@@ -138,7 +138,7 @@ def exclude_filename(f: str):
 
 def go():
     fail = False
-    copyright_pattern = re.compile(r"Copyright 20(18|19|20) The GraphicsFuzz Project Authors")
+    copyright_pattern = re.compile(r"Copyright 20(18|19|20|21) The GraphicsFuzz Project Authors")
     generated_pattern = re.compile(r"(g|G)enerated")
 
     for (dirpath, dirnames, filenames) in os.walk(os.curdir, topdown=True):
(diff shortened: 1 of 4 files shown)


paulthomson (approved):
LGTM!

Issue #1016, opened 2020-06-25
Replace integer built-ins with calls to functions that implement those built-ins


afd:
Rewrite e.g. abs as a call to a function, _GLF_abs, that does the absolute value computation.  Similar for mod, min, max, clamp and possibly other functions.

sneh19337:
Hello, can I work on this issue?

shubhav99:
Hi there 
I want to work on this issue 
Please guide me further

sneh19337:
I am already working on this.

shubhav99:
ok sir 
np

afd:
Thanks both for your interest. @sneh19337 do let us know if you've got questions about how to proceed with this issue. You might like to look at the advice I posted on #958.

sneh19337:
Ok sir

sneh19337:
Hello sir, I was trying to do a PR and the following error is shown-
Error:  Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (1-check-style) on project fuzzerserver: You have 3 Checkstyle violations. -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :fuzzerserver
Error: Process completed with exit code 1.

Please suggest me some way to solve this issue

sneh19337:
Sir,
import com.graphicsfuzz.server.thrift.FuzzerService;
import com.graphicsfuzz.server.thrift.FuzzerServiceManager;
import com.graphicsfuzz.server.thrift.GetWorkerNameResult;
import com.graphicsfuzz.server.thrift.ImageJob;
import com.graphicsfuzz.server.thrift.ImageJobResult;
import com.graphicsfuzz.server.thrift.Job;
import com.graphicsfuzz.server.thrift.JobStatus;
import com.graphicsfuzz.server.thrift.WorkerNameError;
These imports are showing errors. There are no maven dependencies for these also. Please tell some way to solve this problem.

KishkinJ10:
i am a beginner can i work on this ??

sneh19337:
[...]

PR #1126, opened 2021-03-23
test

sneh19337:
Through this PR I have tried to resolve issue #1016 . I have written self-implemented functions in place of integer built-in like max,min,sqrt, abs,ceil.


sneh19337:
Is this some kind of server issue. Or some issue from my side. Please check and tell.

PR #1153, opened 2021-07-30
Add support for std140 qualifier

AaronGhost:
Updates the AST builder and adds a class for the qualifier.


ast/src/main/java/com/graphicsfuzz/common/ast/type/Std140LayoutQualifier.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2019 The GraphicsFuzz Project Authors
```
afd, line 2: Please change copyright year to 2021.

afd (changes requested):
Minor change requested.

ast/src/main/java/com/graphicsfuzz/common/ast/type/Std140LayoutQualifier.java
AaronGhost: Done

afd (approved):

PR #1154, opened 2021-07-30
Update of the typer with shifts and assignments

AaronGhost:
Updates the Typer class to produce a result in the case of assignments
and the TyperHelper class to correctly type shift operations in the
case of left and right scalar types (the resulting type is the one
from the left operand).

### DIFF
--- a/ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
+++ b/ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
@@ -58,10 +58,12 @@ public final class TyperHelper {
       return lhsType;
     }
 
-    // If one side is scalar and the other side is basic, the result has to be that of the other
-    // side
-    // If both are scalar but of different types (shift operator), the result type is the one
-    // from the left operand, so we check if the right operand is scalar first
+    // If one side is scalar, the other side is either a scalar of different type or a non-scalar
+    // basic type:
+    // - If the two operands are scalar of different types, the operation is a shift and
+    // the result type is the type of the left operand
+    // - If one operand is scalar and the other is basic, the result type is the one of the basic
+    // type
     if (rhsType == BasicType.FLOAT || rhsType == BasicType.INT || rhsType == BasicType.UINT) {
       assert lhsType instanceof BasicType;
       return lhsType;
@@ -71,7 +73,8 @@ public final class TyperHelper {
       return rhsType;
     }
     // Now we are in a position where if we know that one type is vector
-    // or matrix, the other side must be also
+    // or matrix, the other side must be also and the result type is the one from the type on the
+    // left if the two types are different (shift operations)
     for (BasicType t : Arrays.asList(
         BasicType.VEC2,
         BasicType.VEC3,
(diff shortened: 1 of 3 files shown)


ast/src/main/java/com/graphicsfuzz/common/typing/TyperHelper.java
```diff
@@ -60,14 +60,16 @@ public static Type resolveTypeOfCommonBinary(Type lhsType, Type rhsType) {
 
     // If one side is scalar and the other side is basic, the result has to be that of the other
     // side
-    if (lhsType == BasicType.FLOAT || lhsType == BasicType.INT || lhsType == BasicType.UINT) {
-      assert rhsType instanceof BasicType;
-      return rhsType;
-    }
+    // If both are scalar but of different types (shift operator), the result type is the one
+    // from the left operand, so we check if the right operand is scalar first
```
afd, line 64: Can you clean up this comment a bit? The existing commentary potentially contradicts the new commentary. If one side is int and the other is uint, the first part of the commentary is ambiguous. And actually if the LHS is ivec2 but the RHS is uint, the result is ivec2 (in the case of a shift), contradicting the first part of the comment.

Feel free to simply rethink this whole section of comment to write something from scratch that reflects our new understanding.

afd (changes requested):
Thanks for spotting this discrepancy!

Please add a test along these lines to TyperTest, right after `testLengthTyped`:

```
  @Test
  public void testLengthTyped() throws Exception {
    final TranslationUnit tu = ParseHelper.parse("#version 320 es\n"
        + "void main() {\n"
        + "  int a;\n"
        + "  uint b;\n"
        + "  ivec2 av;\n"
        + "  uvec2 bv;\n"
        + "\n"
        + "  int x0 = a << a;\n"
        + "  int x1 = a << b;\n"
        + "  uint x2 = b << a;\n"
        + "  uint x3 = b << b;\n"
        + "  ivec2 x4 = av << a;\n"
        + "  ivec2 x5 = av << b;\n"
        + "  ivec2 x6 = av << av;\n"
        + "  ivec2 x7 = av << bv;\n"
        + "  uvec2 x8 = bv << a;\n"
        + "  uvec2 x9 = bv << b;\n"
        + "  uvec2 x10 = bv << av;\n"
        + "  uvec2 x11 = bv << bv;\n"
        + "}\n");
    new NullCheckTyper(tu) {
      private int counter = 0;
      @Override
      public void visitBinaryExpr(BinaryExpr binaryExpr) {
        super.visitBinaryExpr(binaryExpr);
        if (binaryExpr.getOp() == BinOp.LEFT_SHIFT) {
          switch(counter) {
            case 0:
            case 1:
            assertSame(BasicType.INT, lookupType(binaryExpr));
            break;
            case 2:
            case 3:
            assertSame(BasicType.UINT, lookupType(binaryExpr));
            break;
            case 4:
            case 5:
            case 6:
            case 7:
            assertSame(BasicType.IVEC2, lookupType(binaryExpr));
            break;
            case 8:
            case 9:
            case 10:
            case 11:
            assertSame(BasicType.UVEC2, lookupType(binaryExpr));
            break;
            defalt:
            fail();
          }
          counter++;
        }
      }
    }.visit(tu);
  }
```

It's probably syntactically incorrect but hopefully you get the gist and can fix it up (and fix your code to make it pass, if needed).

And can you adapt its name suitably for the << operator, and add an analogous test for the >> operator?

Plus there's one comment in the review about a comment.

AaronGhost:
Done!

afd (approved):
LGTM.

PR #1155, opened 2021-07-30
Add new basic type groups returning functions

AaronGhost:
Adds new utility functions to return lists of all signed integer based
types (int, ivec2, ivec3, ivec4) and changes the availability of the
boolean utility function.


afd (approved):
LGTM!

PR #1159, opened 2021-09-11
Allow restricting to WGSL-supported builtins

afd:
GraphicsFuzz can be used to test WGSL implementations, via translators
from GLSL to WGSL (possibly via SPIR-V). But this only works for shaders
that are WGSL compatible. This change adds an option to control this,
and avoids generating a number of builtins that are not currently
supported in the WGSL spec.


paulthomson (approved):
LGTM!

PR #1164, opened 2021-09-17
gfauto: update signature_util.py

paulthomson:
* Fix PATTERN_ANDROID_BACKTRACE_CATCHALL to include hex address.
* Add Mali error regex.
* Add tests for the above.


gfauto/gfauto/signature_util.py
```diff
@@ -221,6 +225,17 @@ def get_signature_from_log_contents(  # pylint: disable=too-many-return-statemen
     # noinspection PyUnusedLocal
     group: Optional[str]
 
+    # mali error.
```
afd, line 228: mali-> Mali

afd (approved):
LGTM with one nit.

gfauto/gfauto/signature_util.py
paulthomson: Fixed.

PR #1166, opened 2021-09-29
Update 2021.md

r0ushann:
fixed  some grammatical errors  and added even lines wherever required.


afd:
I didn't check them all, but I don't agree with the grammatical changes that I did look at, and also this file is historical and we don't want to change it.

PR #1169, opened 2021-12-07
Apply floor() to gl_FragCoord

jarisiru:
This change applies floor() to the uses of gl_FragCoord in order to
eliminate the potential ambiguity of having 0.5 in the source values.
This should improve the stability of the shaders in fuzzing.


paulthomson (commented):
Thanks for doing this. I would like to hear Ally's opinion. Some initial thoughts: maybe we should only do this for stable_ shaders. Also, if there are statements like gl_FragCoord.x < ..., then I think we can (and should) avoid flooring in this case. Also, I wonder if doing `- 0.5` instead of `floor` might trigger different compiler behavior; not sure which is best though. I suppose we could do a mix. I think the use of gl_FragCoord.x in the gfauto texture generation code is possibly the most worrying because the results are stored into an 8 bit value, but of course we can address that in a different PR.

Issue #1179, opened 2022-01-10
glsl-reduce enhancement: based on perses


AaronGhost:
I have been experimenting with both glsl-reduce and perses as reducers. It seems that Perses performs more reduction opportunities than glsl-reduce, especially in reducing more aggressively array indexes. Please find below some examples of structures reduced first with glsl-reduce and then with perses.

- [ ] Reducing ternary operators in array indexes

**glsl-reduce**
``` glsl
void main()
{
 ext_2[(all(bvec2(true)) ? (+ abs(ext_1[1])) : (ext_0 & -- ext_2[2]))] |= 1;
}
```

**perses**
``` glsl
void main()
{
 ext_2[abs(ext_1[1])] |= 1;
}
```

- [ ]  Removing the outer array in an array index expression

**glsl-reduce**
``` glsl
void main()
{
 ext_3[(+ + (ext_4 /= (ext_5[2086006349])))] = 1u;
 ++ ext_4[int(var_0[ext_4[(ext_4[ext_4[(ext_4[ext_4[ext_4[1893896488]]])]])]].ttt)];
}
```

**perses**
``` glsl
void main()
{
 ext_4 /=  ext_5[2086006349];
 ++ ext_4[int(var_0[ext_4[1893896488]])];
}
```

- [ ] Reducing complicated swizzles (some can be removed altogether)

**glsl-reduce**
```
void main()
{
  ivec4 var_1 = ivec4(- 2147483634);
 (ext_3[ext_7]) = bitfieldExtract(114u, 1, (~ ext_1) / var_1.ttp.yx.x);
}
```

**perses**
``` glsl
void main()
{
  ivec4 var_1 = ivec4(- 2147483634);
  ext_3[ext_7]  = bitfieldExtract(114u, 1,  ~ ext_1  / var_1.x);
}
```

- [ ]  Reducing the number of parameters in type constructors

**glsl-reduce**
``` glsl
void main()
{
 while(1u + ext_3 <= uint(uvec4(bvec2(false), 1u, true)))
  {
   ext_10 = 1u;
  }
}
```

**perses**
``` glsl
 void main()
{
 while(1u + ext_3 <= uint(false))
 {
   ext_10 = 1u;
 }
}
```

afd:
@AaronGhost For the second one, "Removing the outer array in an array index expression", are you able to paste the necessary buffer block declarations?

PR #1186, opened 2022-03-09
Support for swizzles in the reducer

afd:
Adds support for simplifying, shortening and removing swizzles.

Part of #1179.

### DIFF
--- a/ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
+++ b/ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
@@ -32,7 +32,6 @@ import com.graphicsfuzz.common.ast.expr.ArrayIndexExpr;
 import com.graphicsfuzz.common.ast.expr.BinOp;
 import com.graphicsfuzz.common.ast.expr.BinaryExpr;
 import com.graphicsfuzz.common.ast.expr.BoolConstantExpr;
-import com.graphicsfuzz.common.ast.expr.Expr;
 import com.graphicsfuzz.common.ast.expr.FloatConstantExpr;
 import com.graphicsfuzz.common.ast.expr.FunctionCallExpr;
 import com.graphicsfuzz.common.ast.expr.IntConstantExpr;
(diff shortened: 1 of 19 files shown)


ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
```diff
@@ -1000,6 +1000,56 @@ public void visitBinaryExpr(BinaryExpr binaryExpr) {
...
+        + "void main() {\n"
+        + "  float f;\n"
+        + "  vec3 fv;\n"
+        + "  int i;\n"
+        + "  ivec3 iv;\n"
+        + "  uint u;\n"
+        + "  uvec3 uv;\n"
+        + "  bool b;\n"
+        + "  bvec3 bv;\n"
+        + "  f.x;\n"
```
hevrard, line 1015: Should also test stuff like `f.xxxx`, or `f.xx.xx` ?

ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
```diff
@@ -1000,6 +1000,56 @@ public void visitBinaryExpr(BinaryExpr binaryExpr) {
     }.visit(tu);
   }
 
+  @Test
+  public void testScalarSwizzleTyped() throws Exception {
+    final TranslationUnit tu = ParseHelper.parse("#version 420\n"
+        + "void main() {\n"
+        + "  float f;\n"
+        + "  vec3 fv;\n"
```
hevrard, line 1008: Should we also test vec4, and all swizzle letters `xyzw`, `rgba` and `stpq` ?

hevrard (approved):

ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
afd: We perhaps could, but I actually only added this test to double check that GraphicsFuzz already did the right thing here (which it does) - I considered then removing the test, since I normally write tests for missing features, but thought I may as well leave it in. Thus I won't make it exhaustive.

ast/src/test/java/com/graphicsfuzz/common/typing/TyperTest.java
afd: Similar.

